PR #18, opened 2023-03-05
Update honoWithConvex.ts

sshader:
* Add doc comments to all the things
* Update router to handle `ALL` -- I previously wanted to skip this because it's non-trivial to get our Convex router to support `ALL` (because it's responsible for detecting conflicts). But it's much easier to just get the `HttpRouterWithHono` to support `ALL`


convex/lib/honoWithConvex.ts
```diff
@@ -1,15 +1,56 @@
...
+} from "convex/server";
 import { Hono } from "hono";
 import { httpEndpoint, HttpEndpointCtx } from "../_generated/server";
 
-// Idk where this type is supposed to come from but hack to get it working
+/**
+ * Hono uses the `FetchEvent` type internally, which has to do with service workers
+ * and isn't included in the Convex tsconfig.
+ *
+ * As a workaround, define this type here so Hono + Convex compiles.
```
sshader, line 14: Some googling indicates that I could add `webworker` to my `tsconfig.json` under `lib`, but this is incompatible with `dom` (which is what our default `convex/tsconfig.json` uses), so sticking with this hack but with a better comment

convex/lib/honoWithConvex.ts
```diff
@@ -26,35 +67,65 @@ export class HttpRouterWithHono extends HttpRouter {
...
   }
 
-  // Used to populate the list of routes on the Functions page of the dashboard
+  /**
+   * Returns a list of routed HTTP endpoints.
+   *
+   * These are used to populate the list of routes shown in the Functions page of the Convex dashboard.
+   *
+   * @returns - an array of [path, method, endpoint] tuples.
+   */
```
sshader, line 76: Ported the comments from `HttpRouter` over

convex/lib/honoWithConvex.ts
```diff
@@ -26,35 +67,65 @@ export class HttpRouterWithHono extends HttpRouter {
...
+   * http.lookup("/profile/abc", "GET") // returns [getProfile, "GET", "/profile/*"]
+   *```
+   *
+   * @returns - a tuple [PublicHttpEndpoint, method, path] or null.
+   */
   lookup = (path: string, method: RoutableMethod | "HEAD") => {
     const match = this._app.router.match(method, path);
     if (match === null) {
-      return [this._handler, method as RoutableMethod, path] as const;
+      return null;
```
sshader, line 128: I forgot that these could just return `null` when there was no match 🤦‍♀️

sshader:
(I know I can just merge, but requesting a review in case there are edits to comments I can make without having to add a separate PR / commit)

convex/lib/honoWithConvex.ts
ianmacartney: does that preclude the NotFound handling?

convex/lib/honoWithConvex.ts
```diff
@@ -1,15 +1,56 @@
-import { HttpRouter, PublicHttpEndpoint, RoutableMethod } from "convex/server";
+import {
```
ianmacartney, line 1:
```suggestion
/**
 * Use these helpers to use Hono for routing and middleware in Convex.
 * In `convex/http.ts`:
 * ```ts
 * const app: HonoWithConvex = new Hono();
 * ... use app to define routes
 * export default new HttpRouterWithHono(app);
 * ```
 */
import {
```

convex/lib/honoWithConvex.ts
sshader: Ah good point -- if you have middleware on something like `*`, then we get Hono's 404. If you don't, then we get the Convex one. Seems preferable to just always get the Hono 404 so updated.

PR #22, opened 2023-04-19
Attempt to upgrade to 0.13.0

sshader:
I really just wanted to upgrade the Hono one, but took a stab at some of the other stuff.

The session types were... rough, but it at least compiles and can be a starting place if I didn't do this right.


convex/lib/withZod.ts
```diff
@@ -10,143 +10,63 @@ import {
 import { Id, TableNames } from "../_generated/dataModel";
 
 /**
- * Zod helper for a Convex Id, used for validation.
+ * Create a validator for a Convex `Id`.
```
sshader, line 13: I copied this from the convex monorepo, so lmk if we don't want all these changes

src/hooks/session.ts
```diff
@@ -92,17 +93,13 @@ type ValidQueryNames = {
...
-  ...args: SessionFunctionArgs<NamedQuery<API, Name>>
+  args: SessionFunctionArgs<NamedQuery<API, Name>>
 ) => {
   const sessionId = useContext(SessionContext);
-  // I'm sorry about this. We know that ...args are the arguments following
-  // a Id<"sessions"> | null, so it should always work. It's hard for typescript,
-  // go easy on the poor little inference machine. Also open to ideas about how
-  // to do this correctly.
-  const newArgs = [sessionId, ...args] as unknown as Parameters<
+  const newArgs = { ...args, sessionId } as Parameters<
```
ianmacartney, line 99: wow I'm so happy about this

src/hooks/session.ts
```diff
@@ -122,15 +119,15 @@ export const useSessionMutation = <Name extends ValidMutationNames>(
...
   return (
-    ...args: SessionFunctionArgs<NamedMutation<API, Name>>
+    args: SessionFunctionArgs<NamedMutation<API, Name>>
   ): Promise<ReturnType<NamedMutation<API, Name>>> => {
     // I'm sorry about this. We know that ...args are the arguments following
     // a Id<"sessions"> | null, so it should always work. It's hard for typescript,
     // go easy on the poor little inference machine. Also open to ideas about how
     // to do this correctly.
-    const newArgs = [sessionId, ...args] as unknown as Parameters<
+    const newArgs = { ...args, sessionId } as unknown as Parameters<
```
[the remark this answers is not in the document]
ianmacartney: Something like `{ ...args || {}, sessionId }`? The typing I'm not sure about. And maybe args is already becoming an empty object automatically?

ianmacartney (approved):
tysm! Will think more on typing sessions helpers

PR #25, opened 2023-05-12
add RowLevelSecurity middleware

ldanilek:
add middleware helper for implementing row-level-security.

as far as i can tell, the typescript types work for all ways of defining queries and mutations.

we can discuss on this PR or on the blog post draft or in person.

### DIFF
--- a/convex/lib/rowLevelSecurity.ts
+++ b/convex/lib/rowLevelSecurity.ts
@@ -1,0 +1,404 @@
+import {
+  DatabaseReader,
+  DatabaseWriter,
+  DocumentByInfo,
+  DocumentByName,
+  Expression,
+  FilterBuilder,
+  FunctionArgs,
+  GenericDataModel,
+  GenericTableInfo,
+  IndexRange,
+  IndexRangeBuilder,
+  Indexes,
+  MutationCtx,
+  NamedIndex,
+  NamedSearchIndex,
+  NamedTableInfo,
+  OrderedQuery,
+  PaginationOptions,
+  PaginationResult,
+  Query,
+  QueryCtx,
+  QueryInitializer,
+  SearchFilter,
+  SearchFilterBuilder,
+  SearchIndexes,
+  TableNamesInDataModel,
+} from "convex/server";
+import { GenericId } from "convex/values";
+
+export type Rule<
+  Ctx,
+  DataModel extends GenericDataModel,
+  TableName extends TableNamesInDataModel<DataModel>
+> = (
+  ctx: Ctx,
+  message: DocumentByName<DataModel, TableName>
+) => Promise<boolean>;
+
+export type Rules<Ctx, DataModel extends GenericDataModel> = {
+  [T in TableNamesInDataModel<DataModel>]?: Rule<Ctx, DataModel, T>;
+};
+
+/**
+ * Apply row level security (RLS) to queries and mutations with the returned
+ * middleware functions.
+ *
+ * Example:
+ * ```
+ * // Defined in a common file so it can be used by all queries and mutations.
+ * const {withMutationRLS} = RowLevelSecurity<{auth: Auth, db: DatabaseReader}, DataModel>(
+ *  {
+ *    cookies: async ({auth}, cookie) => !cookie.eaten,
+ *  },
+ *  {
+ *    cookies: async ({auth, db}, cookie) => {
+ *      const user = await getUser(auth, db);
+ *      return user.isParent;  // only parents can reach the cookies.
+ *    }
+ *  }
+ * );
+ * // Mutation with row level security enabled.
+ * export const eatCookie = mutation(withMutationRLS(
+ *  async ({db}, {cookieId}) => {
+ *   // throws "does not exist" error if cookie is already eaten or doesn't exist.
+ *   // throws "write access" error if authorized user is not a parent.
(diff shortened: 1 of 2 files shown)


convex/lib/rowLevelSecurity.ts
```diff
@@ -0,0 +1,404 @@
...
+ * middleware functions.
+ *
+ * Example:
+ * ```
+ * // Defined in a common file so it can be used by all queries and mutations.
+ * const {withMutationRLS} = RowLevelSecurity<{auth: Auth, db: DatabaseReader}, DataModel>(
+ *  {
+ *    cookies: async ({auth}, cookie) => !cookie.eaten,
+ *  },
+ *  {
```
ianmacartney, line 55: open to thoughts, but it might be clearer to take one arg: 
```
{
  read: { cookies: ...},
  write: {cookies: ...}
}
```
Also I know the motivation for making write optional vs. having a merged list, but I imagine wanting to define the rules for each table next to each other. we could do the splitting inside RowLevelSecurity so you don't have to rewrite the wrappers? e.g.:
```
{
  cookies: {
    read: ...
    write: ...
  },
  ...
}
```

convex/lib/rowLevelSecurity.ts
```diff
@@ -0,0 +1,404 @@
...
+ *  }
+ * );
+ * export default query(withUser(withRLS(...)));
+ * ```
+ *
+ * @param readAccessRules - rule for each table, determining whether a row
+ *  should be visible.
+ * @param writeAccessRules - rule for each table, determining whether a row
+ *  may be written by `patch`, `replace`, or `delete`.
+ *  Rules do not restrict `db.insert`.
```
ianmacartney, line 90: so any user can create rows? can't we pass the new document in? Add _creationTime as Date.now(), and _id from the insert? Let's make sure to call this out as a caveat in the readme. in the future we could also extend like:
```
{
  cookies: {
    read: ...
    write: ...
    validateInserts: true
  },
  ...
}
```
where your validation function will take in the null on inserts, if validating the new object doesn't make sense.

convex/lib/rowLevelSecurity.ts
```diff
@@ -0,0 +1,404 @@
...
+  }
+  async insert<TableName extends string>(
+    table: TableName,
+    value: any
+  ): Promise<any> {
+    // No auth check on insert.
+    return await this.db.insert(table, value);
+  }
+  async checkAuth<TableName extends string>(id: GenericId<TableName>) {
+    const doc = await this.get(id);
```
ianmacartney, line 372: Let's also make it clear that all operations will now do reads too - so a batch delete function should definitely try to parallelize, etc! It's actually so nice that our writer API was already async - this wouldn't be possible with the synchronous writing API
I've in the past done `patch` instead of get & replace b/c I thought it'd be faster. In this world, it might be worth mentioning that it won't be, you may as well grab & replace.

ianmacartney (commented):
Let's add a section in README.md pointing to this and (optimistically based on the article slug we'll use - https://stack.convex.dev/row-level-security perhaps?) pointing to the stack article

convex/lib/rowLevelSecurity.ts
ianmacartney: or any made up _id - ideally the auth check shouldn't bleed info out, and the _id doesn't exist yet so they don't have much to check. or we pass in the `withoutSystemFields` type and hope they don't need the _id? Or make _id optional so they have to check in their authorization fn, and `null` means it hasn't been inserted yet

ldanilek: how about a separate validation rule for inserts that can take in a `WithoutSystemFields`

convex/lib/rowLevelSecurity.ts
ldanilek: agree that is nicer

convex/lib/rowLevelSecurity.ts
ldanilek: i don't think there's much of a difference here. left a comment to that effect

convex/lib/rowLevelSecurity.ts
```diff
@@ -48,7 +46,12 @@ export type Rules<Ctx, DataModel extends GenericDataModel> = {
  * Example:
  * ```
  * // Defined in a common file so it can be used by all queries and mutations.
- * const {withMutationRLS} = RowLevelSecurity<{auth: Auth, db: DatabaseReader}, DataModel>(
+ * import { Auth } from "convex/server";
+ * import { DataModel } from "./_generated/dataModel";
+ * import { DatabaseReader } from "./_generated/server";
+ * import { RowLevelSecurity } from "./rowLevelSecurity";
+ * 
+ * export const {withMutationRLS} = RowLevelSecurity<{auth: Auth, db: DatabaseReader}, DataModel>(
```
ianmacartney, line 54: let's update the comment example to match the new format

convex/lib/rowLevelSecurity.ts
ianmacartney: how about read / write / modify, where modify = "write" and write = "insert"?
"write" seems more like insert. or just change "write" to "modify" and leave insert?

ianmacartney (approved):
woohoo! Very excited

PR #31, opened 2023-06-14
update to convex 0.17.0

ldanilek:
this is a draft because 0.17.0 is still in alpha.
mostly wanted to send this because the rowLevelSecurity changes are working.

i haven't tested the changes to anything besides rowLevelSecurity, just that the convex/ directory compiles.

i expect we will need to pick something to do for `withZod`, and also migrate the custom hooks to use `api....` syntax, and fix the migration action that passes a mutation reference to an action.


convex/lib/withZod.ts
```diff
@@ -16,9 +16,7 @@ import { Id, TableNames } from "../_generated/dataModel";
  * @returns - A Zod object representing a Convex `Id`
  */
 export const zid = <TableName extends TableNames>(tableName: TableName) =>
```
sshader, line 18: I feel like we should just not support `zid` as a custom validator at all because it gives a false sense of security.  E.g. "foo" would validate as a `Id<"users">` because it's a string.

Curious how we feel about pushing folks towards `v.id("users")` instead, which will do the right thing.

ianmacartney: When you say "push folks toward v.id" do you mean deprecate using Zod? Or is there a way to use `v.id` separately?
The goal of `withZod` is to give users who already use Zod or prefer it, to use it as a drop-in replacement for our `{args, handler}` validation. 
One amazing extension of this would be to translate zod types to convex `args` validation and **also** then do follow-on Zod validation. So Convex could check basic types, and Zod could do any additional things like phone numbers, etc. That would also solve the problem of wanting to use Zod and also wanting the dashboard/ other languages to be able to know the argument shape.
I know @thomasballinger (and others?) talked with the Zod folks about this being possible already.
Then the challenge would just be how to have a "convex ID" string that turns to a string / passthrough validator in Zod and turns into `v.id` internally...

convex/lib/rowLevelSecurity.ts
```diff
@@ -365,6 +391,16 @@ class WrapWriter<Ctx, DataModel extends GenericDataModel>
...
     return await this.db.insert(table, value);
   }
+  tableName<TableName extends string>(
+    id: GenericId<TableName>
+  ): TableName | null {
+    for (const tableName of Object.keys(this.rules)) {
+      if (this.db.normalizeId(tableName, id)) {
+        return tableName as TableName;
+      }
+    }
```
ianmacartney, line 401: This makes me nervous to suggest this as a serious solution for folks looking for RLS. Between this and Zod, double checking: it's not possible to expose a `getTableName` non-db function only callable from the server @sshader?

convex/lib/withZod.ts
ianmacartney: If we don't make this work, it's pretty inconvenient to not have something that ends up typed as Id<"tablename>": that will show up as a string param on the client and in the function for every Id...

PR #32, opened 2023-06-19
Update convex-helpers to 0.17.0

sshader:
Built off of https://github.com/get-convex/convex-helpers/pull/31

I also worked in some middleware helpers I had been using in our own repo. Our existing ones had a bug where if you had a function with args but no validator, the session middleware would fail because it would have a partial validator for the session ID.

I also changed the `ConvexReactClientFake` to use `registerMutationFake`. The test mocking the import doesn't seem to pass and I'm not really sure why...


convex/lib/withSession.ts
```diff
@@ -34,47 +39,29 @@ const optionalSessionIdValidator = v.union(v.id("sessions"), v.null());
...
-  };
-}
+const optionalSessionMiddlewareValidator = {
+  sessionId: v.union(v.null(), v.id("sessions")),
+};
+const transformContextForOptionalSession = async <Ctx>(
+  ctx: Ctx & { db: DatabaseReader },
+  args: ObjectType<typeof optionalSessionMiddlewareValidator>
+): Promise<Ctx & { session: Doc<"sessions"> | null }> => {
+  const session = args.sessionId ? await ctx.db.get(args.sessionId) : null;
```
ldanilek, line 49: can we do this to be more bulletproof, especially for people upgrading to 0.17.0 for an existing app
```suggestion
  const sessionId = args.sessionId ? ctx.db.normalizeId("sessions", args.sessionId) : null;
  const session = sessionId ? await ctx.db.get(args.sessionId) : null;
```

sshader: Hmm I wonder if we should do a separate function for people upgrading then. This change alone wouldn't be enough (the validator would fail first), and there's other things one might want for upgrading (e.g. writing back the updated value to local storage / session storage)

convex/presence.ts
```diff
@@ -1,12 +1,3 @@
-/**
- * Functions related to reading & writing presence data.
- *
- * Note: this file does not currently implement authorization.
- * That is left as an exercise to the reader. Some suggestions for a production
- * app:
- * - Use Convex `auth` to authenticate users rather than passing up a "user"
- * - Check that the user is allowed to be in a given room.
- */
```
ianmacartney, line 9: why snip?

convex/lib/withUser.ts
```diff
@@ -1,21 +1,3 @@
...
- * ```
- * mutation(withUser(({db, user}) => {...})) // fails to infer that "db" is a DatabaseWriter
- * mutation(withUser({ handler: ({db, user}) => {...} })) // Works!
- * mutationWithUser({ handler: ({db, user}) => {...} })   // Works!
- * mutationWithUser(({db, user}) => {...})                // Works!
- * mutation(withUser(
- * ({db, user}: MutationCtx & {user: Doc<"users">} ) => {...}
- * ))                                                     // Works!
- * ```
- */
```
ianmacartney, line 18: why snip?

convex/lib/withSession.ts
```diff
@@ -203,22 +215,26 @@ export function queryWithSession<
...
   >
-): RegisteredQuery<"public", [{ sessionId: Id<"sessions"> | null }], Output>;
+): RegisteredQuery<
+  "public",
+  MergeArgsForRegistered<Args, ObjectType<typeof sessionMiddlewareValidator>>,
+  Output
+>;
 export function queryWithSession(func: any): any {
-  return query(withOptionalSession(func));
+  return query(withSession(func));
```
ianmacartney, line 239: query is no longer optional by default? I found, without adding an app-wide spinner for sessions, that session queries were the ones most likely to need to handle un-initialized sessions

ianmacartney (commented):
haven't looked at it all yet, but some thoughts to start

convex/lib/withUser.ts
sshader: oops -- I think this got stripped away by the codemod from `"messages:list"` to `api.messages.list` when modifying imports. will add them all back

thomasballinger (approved):
This looks great, I think it makes writing middleware less intimidating.

Side note, playing with the branch I'm reminded it'd be nice to have some tests and examples in the repo.

thomasballinger:
@ianmacartney I see you said "some thoughts to start," is anything you have blocking?

PR #36, opened 2023-07-06
allow empty arg for useSessionMutation too

ianmacartney:
I just noticed that the session middleware now requires you pass {} as an arg if the session ID was the only arg.
useSessionQuery spreads the args. Any reason mutation shouldn't?


sshader (approved):
Oops I totally thought I had approved this already. Seems good (tried it out in our monorepo where we have slightly more test cases). But also cc @thomasballinger who's played around with the client-side hooks more than I have.

thomasballinger:
👍 no thoughts, sounds good

PR #41, opened 2023-12-09
Implement customFunction to replace middlewareUtils

ianmacartney:


convex/lib/withSession.ts
```diff
@@ -125,80 +38,48 @@ export const withOptionalSession = generateMiddleware<
...
-  MergeArgsForRegistered<Args, ObjectType<typeof sessionMiddlewareValidator>>,
-  Output
->;
-export function mutationWithSession(func: any): any {
-  return mutation(withSession(func));
-}
+export const mutationWithSession = customMutation(mutation, {
+  args: { sessionId: v.id("sessions") },
+  input: async (ctx, args) => {
+    const session = (await ctx.db.get(args.sessionId)) ?? null;
```
sshader, line 44: Is it intentional that mutation requires a session but query does not?

convex/lib/withSession.ts
```diff
@@ -125,80 +38,48 @@ export const withOptionalSession = generateMiddleware<
...
+    if (session === null) {
+      throw new Error(
+        "Session must be initialized first. " +
+          "Are you wrapping your code with <SessionProvider>? " +
+          "Are you requiring a session from a query that executes immediately?"
```
sshader, line 49: This shouldn't say `query`, right? Maybe remove the line?

sshader (commented):
I would be excited to see the READMEs / docs for how to use `customFunction` (could also potentially take a stab at this myself)

ianmacartney:
> I would be excited to see the READMEs / docs for how to use customFunction (could also potentially take a stab at this myself)

README in progress!

PR #54, opened 2024-01-08
Add CustomCtx type helper

xixixao:
I'm not sure how to test this exactly in this repo, but I tested it in mine and it works.

In my file with custom functions I can now do:

```ts
export const queryFoo = customQuery(
  baseQuery,
  customCtx(async (ctx) => {
    return {
      /// some customization....
    };
  })
);

export type QueryFooCtx = CustomCtx<typeof query>;
```


ianmacartney (approved):
love it, thanks for adding this

PR #56, opened 2024-01-18
Add index to async map asyncTransform params

Terris:
I found myself wanting `asyncMap` to provide an index in `asyncTransform` props.

In my case I was updating an order column on a list of records and found it convenient to have an index while mapping over them and building a return. For instance: 

```javascript
// Update the order of all courseLessons to keep
await asyncMap(courseLessons, (cm, index) => {
   return ctx.db.patch(cm._id, { order: index + 1 });
});
```

So, this PR adds an index to the asyncTransform params on asyncMap! Might be a rare use case, but I thought I'd put it out there anyhow.

```javascript
/**
 * asyncMap returns the results of applying an async function over an list.
 *
 * @param list - Iterable object of items, e.g. an Array, Set, Object.keys
 * @param asyncTransform
 * @returns
 */
export async function asyncMap<FromType, ToType>(
  list: Iterable<FromType>,
  asyncTransform: (item: FromType, index: number) => Promise<ToType> // add index and type to param
): Promise<ToType[]> {
  const promises: Promise<ToType>[] = [];
  let index = 0; // init an index var
  for (const item of list) {
    promises.push(asyncTransform(item, index)); // return the index with the item
  }
  return Promise.all(promises);
}
```

Thanks for considering :)

### DIFF
--- a/packages/convex-helpers/index.ts
+++ b/packages/convex-helpers/index.ts
@@ -13,6 +13,7 @@ export async function asyncMap<FromType, ToType>(
   let index = 0;
   for (const item of list) {
     promises.push(asyncTransform(item, index));
+    index += 1;
   }
   return Promise.all(promises);
 }
--- a/convex/relationshipsExample.ts
+++ b/convex/relationshipsExample.ts
@@ -55,7 +55,7 @@ export const relationshipTest = mutation({
     const edges = await getManyFrom(
       ctx.db,
       "join_table_example",
-      "userId",
+      "by_userId",
       userId
     );
     assertLength(edges, 2);
@@ -63,7 +63,7 @@ export const relationshipTest = mutation({
       ctx.db,
       "join_table_example",
       "sessionId",
-      "userId",
+      "by_userId",
       userId
     );
     assertLength(sessions, 2);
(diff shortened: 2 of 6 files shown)


ianmacartney:
I love that it also mimics the behavior of built-in map. I'll try it out &
hopefully fold this into the next release.
Thank you!

ianmacartney:
I added incrementing the index :P

packages/convex-helpers/index.ts
```diff
@@ -7,11 +7,13 @@
...
   list: Iterable<FromType>,
-  asyncTransform: (item: FromType) => Promise<ToType>
+  asyncTransform: (item: FromType, index: number) => Promise<ToType>
 ): Promise<ToType[]> {
   const promises: Promise<ToType>[] = [];
+  let index = 0;
   for (const item of list) {
-    promises.push(asyncTransform(item));
+    promises.push(asyncTransform(item, index));
+    index += 1;
```
ianmacartney, line 16: other files got added here b/c I pushed the merged main to your branch, but this was the only addition

Terris:
Many thanks, Ian!

ianmacartney:
Released in 0.1.13

PR #67, opened 2024-02-22
CRUD Helpers

ianmacartney:
Helpers to add Create Read Update and Delete functions for a table.

### DIFF
--- a/packages/convex-helpers/server/index.ts
+++ b/packages/convex-helpers/server/index.ts
@@ -1,5 +1,15 @@
-import { defineTable } from "convex/server";
-import { Validator, v } from "convex/values";
+import {
+  defineTable,
+  QueryBuilder,
+  MutationBuilder,
+  GenericDataModel,
+  WithoutSystemFields,
+  DocumentByName,
+  RegisteredMutation,
+  RegisteredQuery,
+  FunctionVisibility,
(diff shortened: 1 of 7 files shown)


packages/convex-helpers/server/index.ts
```diff
@@ -68,3 +79,105 @@ export function deploymentName() {
...
+      args: { id: table._id },
+      handler: async (ctx, args) => {
+        await ctx.db.delete(args.id);
+      },
+    }) as RegisteredMutation<
+      MutationVisibility,
+      { id: GenericId<TableName> },
+      Promise<void>
+    >,
+  };
```
ianmacartney, line 182: I made them all title case b/c I couldn't define `delete`: I considered `delete_` but it didn't feel right. and `remove` doesn't match the D in CRUD

vakila: In templates-cms I used "destroy" for this, just an idea in case you want to stick to lowercase as more in keeping with the rest of examples etc - but I also think title case is OK

vakila: oh I just saw the later commit going back to "delete_", I think personally I'd prefer sticking to consistent naming across all 4 - either all title case or "destroy"

PR #68, opened 2024-02-22
Add Retries helpers and example

ianmacartney:
Adds the retry helpers from https://github.com/JamesCowling/convex-action-retrier
Changes the API to instead of exposing a mutation to run, expose a type-safe function that schedules both, whether you're in an action or mutation. The only risk with this is if you call it from an action and the scheduler succeeds on the first but fails the retrier on the second, and the called action fails, which seems exceedingly unlikely.

### DIFF
--- a/convex/retriesExample.ts
+++ b/convex/retriesExample.ts
@@ -6,7 +6,8 @@ import { makeActionRetrier } from "convex-helpers/server/retries";
 
 // Export the helpers, with the name of the retry function.
 export const { runWithRetries, retry } = makeActionRetrier(
-  "retriesExample:retry"
+  "retriesExample:retry",
+  { retryBackoff: 1000 } // set to 1 second for demo purposes.
 );
 
 // This is a sample action will fail randomly based on the `failureRate`
@@ -22,21 +23,68 @@ export const unreliableAction = internalAction({
   },
 });
 
-// Call this to call the `unreliableAction` function with retries.
+// This calls the `unreliableAction` function with retries.
+// Try it for yourself with:
 // e.g. `npx convex run retriesExample:runUnreliableActionWithRetries`
 export const runUnreliableActionWithRetries = mutation({
-  args: { failureRate: v.optional(v.number()) },
+  args: {},
   handler: async (ctx, args) => {
+    await runWithRetries(ctx, internal.retriesExample.unreliableAction, {
+      failureRate: 0.8,
+    });
+    // Possibly do something else besides scheduling the action,
+    // for instance check if the logged in user has permission to run it.
+    // If an error is thrown in a mutation, the transaction will be rolled back
+    // and the action will not be scheduled to run.
+  },
+});
+
+// Calling an action with retries from an action works too.
+export const runFromAnAction = internalAction({
+  args: {},
+  handler: async (ctx) => {
+    const email = { to: "user@example.com", subject: "Hello", body: "World" };
     await runWithRetries(
       ctx,
-      internal.retriesExample.unreliableAction,
+      internal.retriesExample.sendWelcomeEmail,
+      { email },
       {
-        failureRate: args.failureRate ?? 0.8,
-      },
-      {
-        maxFailures: 4,
-        retryBackoff: 1000,
+        // You can limit the number of retries and wait longer between attempts
+        // if the action is not time sensitive and generally reliable.
+        // Note: generally the default options are fine.
+
+        // After 5 failures, give up.
+        maxFailures: 5,
+        // Wait 10 seconds before retrying the first time.
+        retryBackoff: 10_000,
+        // This is the multiplier for the next wait time.
+        // e.g. if base is 10, the next wait time will be 10 times the previous.
+        base: 10,
+        // Wait time is then retryBackoff * base^(retryNumber).
+        // This will result in waiting:
+        // 10 seconds before retrying the first time,
+        // 100 seconds before retrying the second time (~1.7 minutes),
+        // 1000 seconds before retrying the third time (~16.7 minutes),
+        // 10000 seconds before retrying the fourth and final time (~2.7 hours),
+        // giving the action 5 chances to succeed over ~3 hours.
       }
     );
+    // Unlike in mutations, in an action the scheduler will immediately be
+    // given the action & parameters to run, even if there's an exception thrown
+    // afterwards.
+  },
+});
+
+// This is a pretend email sending function.
+export const sendWelcomeEmail = internalAction({
+  args: {
+    email: v.object({ to: v.string(), subject: v.string(), body: v.string() }),
+  },
+  handler: async () => {
+    // If the email provider is down, it could fail.
+    // Hopefully it will be back up within minutes, but if it still isn't up
+    // hours later, it's ok to skip sending this non-essential.
+    // Note: If we wanted to ensure we don't send the same email twice,
+    // we would need something like an idempotency key to pass to the provider.
   },
 });
--- a/convex/retriesExample.ts
+++ b/convex/retriesExample.ts
@@ -1,0 +1,53 @@
+import { v } from "convex/values";
+import { api, internal } from "./_generated/api";
+import { Doc, Id } from "./_generated/dataModel";
+import {
+  action,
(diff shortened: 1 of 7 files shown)


packages/convex-helpers/server/retries.ts
```diff
@@ -0,0 +1,171 @@
...
+ * // in a mutation or action
+ * await runWithRetries(ctx, internal.myModule.myAction, { arg1: 123 });
+ * ```
+ *
+ * @param internalMutation From "./convex/_generated/server" or customMutation.
+ * @param retryRef The function reference to the retryRef function exported.
+ * e.g. internal.mymodule.retry
+ * @returns An object with runWithRetries and retry functions.
+ */
+export function makeActionRetrier(retryFnName: string) {
```
ianmacartney, line 55: It would be great to have a stable reference to a function. 
Components would work well here - a well-known function to call.
This needs a string instead of the type-safe FunctionReference b/c it'd be circular types.

packages/convex-helpers/server/retries.ts
```diff
@@ -0,0 +1,171 @@
...
+      "action",
+      Visibility,
+      DefaultFunctionArgs,
+      null | Promise<null> | void | Promise<void>
+    >,
+    Visibility extends FunctionVisibility = "internal"
+  >(
+    ctx: { scheduler: Scheduler },
+    action: Action,
+    actionArgs: FunctionArgs<Action>,
```
ianmacartney, line 82: after much back & forth the types for the action arguments weren't too bad (and type safety to ensure passing an action ref instead of a mutation or query which wouldn't make sense).

packages/convex-helpers/server/retries.ts
```diff
@@ -0,0 +1,171 @@
...
+      waitBackoff?: number;
+      retryBackoff?: number;
+      base?: number;
+      maxFailures?: number;
+    }
+  ) {
+    const job = await ctx.scheduler.runAfter(
+      0,
+      action,
+      ...([actionArgs] as OptionalRestArgs<Action>)
```
ianmacartney, line 93: this was some annoying typing - it wouldn't let me just pass them in since it wanted sometimes an EmptyObject. I think this can be improved but haven't dug in

packages/convex-helpers/server/retries.ts
```diff
@@ -0,0 +1,158 @@
...
+    );
+    await ctx.scheduler.runAfter(0, retryRef, {
```
JamesCowling, line 82: If the caller fails at this point the action won't be reliable right?

I think we can't put out a framework for reliable actions that isn't actually reliable. Seems actually quite plausible that the action would fail right here. Is there a reason why we can't have the one scheduled mutation both call the action and retry it?

ianmacartney: tl;dr I think I'll change it to also require a `db: DatabaseWriter` in the ctx, so that it's only callable from mutations to encourage the right behavior. it can be wrapped to call from an action / client.

My thought was that we should be encouraging folks to do this from the server, not expose a public action - public actions can be client-side retried. Or wrap with a mutation.
My reasoning for not putting in its own internalMutation:
A: We don't suggest directly calling mutations from each other, so it won't be transactional with your other data changes when scheduled from a mutation.

B: In an action here, it can fail:

1. before scheduling anything - action will be attempted 0 times
2. between scheduling the action and scheduling the retry - action will be attempted 1 time
3. after scheduling - action will be attempted N times, but further logic won't be attempted.

When you schedule a mutation from an action, it can fail:

1. before scheduling anything - action will be attempted 0 times
2. after scheduling - action will be attempted N times, but further logic won't be attempted.

The incremental "loss" in an action is the possibility that you'd only try something once. 

If we want to avoid this, we should push users to use it transactionally with their other data commits, and force them to wrap it explicitly to call from an action.

Down the line I'd like to roll this sort of behavior into a job mgmt helper where you can subscribe to the job status - how many times its been retried etc. right now you get an ID that is only for the first invocation and can't trace its progress

ianmacartney: I've edited the above, and also realized that since the only "transactional" benefit here is that they are scheduled, and scheduling the wrapper is transactional, there's no transactional benefit. 

**The main reason is that this version has type checking. But we can constrain it to being called by mutations if we want to call it but also ensure it's in a mutation**

convex/retriesExample.ts
```diff
@@ -0,0 +1,44 @@
...
+      throw new Error("action failed.");
+    }
+    console.log("action succeded.");
+  },
+});
+
+// Call this to call the `unreliableAction` function with retries.
+// e.g. `npx convex run retriesExample:runUnreliableActionWithRetries`
+export const runUnreliableActionWithRetries = mutation({
+  args: { failureRate: v.optional(v.number()) },
```
JamesCowling, line 30: nit but i wonder why this is inlined and it's not on line 16. has this run through prettier?
maybe this is just because line 16 has a line comment on it, in which case ignore me

convex/retriesExample.ts
```diff
@@ -0,0 +1,44 @@
...
+  handler: async (ctx, args) => {
+    await runWithRetries(
+      ctx,
+      internal.retriesExample.unreliableAction,
+      {
+        failureRate: args.failureRate ?? 0.8,
+      },
+      {
+        maxFailures: 4,
```
JamesCowling, line 39: Pretty reasonable to increase the default `retryBackoff` to 1s or maybe 100ms (if it's the default) but does `maxFailures: 4` feel a bit low to you in an example? There'll be cases with an 80% failure rate where this just doesn't ever execute successfully.

packages/convex-helpers/server/retries.ts
```diff
@@ -0,0 +1,164 @@
...
+  FunctionArgs,
+  OptionalRestArgs,
+  getFunctionName,
+  makeFunctionReference,
+  DefaultFunctionArgs,
+  internalMutationGeneric,
+} from "convex/server";
+import { v, ObjectType } from "convex/values";
+
+const DEFAULT_WAIT_BACKOFF = 10;
```
JamesCowling, line 19: I mentioned this above but I think 10ms is probably a mistake and we should start with 100ms. Doesn't matter a huge amount but might save them hammering a third party service too much.

If you disagree then just leave it as-is.

packages/convex-helpers/server/retries.ts
```diff
@@ -0,0 +1,164 @@
...
+const DEFAULT_RETRY_BACKOFF = 10;
```
JamesCowling, line 20: Same as above

packages/convex-helpers/server/retries.ts
JamesCowling: I think we shouldn't give the developer an option to ever call this directly from an action. The source code in here is complicated enough that they likely won't really know what's going on and instead will just use it as a black box. So I think we just give them explicit examples of what to do when you want to call a reliable action either from the client, from a mutation, or from another action, the latter which will require scheduling through a mutation.

I think we'll be able to ship once we just are very prescriptive with the example instead of giving them options. This could be as simple as a comment similar to what we have but says explicitly something like "don't call this from an action. schedule a mutation that calls this"

JamesCowling (commented):
Just that one comment about making it really explicit how not to use this. Otherwise all good.

ianmacartney (commented):
I made some structural changes:
- It will always just schedule one mutation, similar to your original approach. So it's no longer unsafe to call from actions.
- It re-uses the retry mutation for this (calling it with an undefined job), so it only has to expose one mutation.
- I outlined examples of calling it from a mutation and action, with the default options, and customizing for fewer retries over a long period of time for heavy workloads.
- I'm not going to document how to call something from the client generically, since I realized that is a pretty big security concern: that would allow the client to invoke any internal function with any arguments; the `makeFunctionReference` function doesn't even let you validate whether the function is an action or mutation.

Once this is landed, I think it's worth updating the retries post a bit:
- show how to use this helper and point to it in the "show me the code" section, then also point to the repo used in the demo for folks to follow along.
- change runAction in the demo repo & post to be an internalMutation so folks don't copy-paste a backdoor.

PR #71, opened 2024-03-04
new filter helper

ldanilek:
create a new function that can wrap any query: `filter` takes in a query and a javascript predicate and uses it for filtering.

i'm thinking this can accompany a stack article that explains that `.filter()` and `filter()` are basically the same.

also changed RowLevelSecurity to use this new filter internally.

added an example

### DIFF
--- a/convex/customFunctionExample.ts
+++ b/convex/customFunctionExample.ts
@@ -13,6 +13,7 @@ import { getUserByTokenIdentifier } from "./lib/withUser";
 import { Rules } from "convex-helpers/server/rowLevelSecurity";
 import { DataModel, Doc } from "./_generated/dataModel";
 import { SessionIdArg, vSessionId } from "convex-helpers/server/sessions";
+import { ReaderWithFilter } from "convex-helpers/server/filterWith.js";
 
 const rules: Rules<{ user: Doc<"users"> }, DataModel> = {
   presence: {
@@ -96,5 +97,26 @@ export const someMutation = myMutationBuilder({
     ctx.sessionId;
     ctx.user;
     return { success: true };
+  },
+});
+
+const queryWithFilter = customQuery(
+  query,
+  customCtx(async (ctx) => {
+    return {
+      db: new ReaderWithFilter(ctx.db),
+    };
+  })
+);
+
+export const queryFiltered = queryWithFilter({
+  args: {},
+  handler: async (ctx) => {
+    return await ctx.db
+      .query("presence")
+      .filterWith(async (presence) => {
+        return presence.updated > 10;
+      })
+      .first();
   },
 });
--- a/convex/customFunctionExample.ts
+++ b/convex/customFunctionExample.ts
@@ -13,7 +13,7 @@ import { getUserByTokenIdentifier } from "./lib/withUser";
 import { Rules } from "convex-helpers/server/rowLevelSecurity";
 import { DataModel, Doc } from "./_generated/dataModel";
 import { SessionIdArg, vSessionId } from "convex-helpers/server/sessions";
-import { ReaderWithFilter } from "convex-helpers/server/filterWith.js";
+import { filter } from "convex-helpers/server/filter";
 
 const rules: Rules<{ user: Doc<"users"> }, DataModel> = {
   presence: {
@@ -100,21 +100,11 @@ export const someMutation = myMutationBuilder({
   },
 });
 
-const queryWithFilter = customQuery(
-  query,
-  customCtx(async (ctx) => {
-    return {
-      db: new ReaderWithFilter(ctx.db),
-    };
-  })
-);
-
-export const queryFiltered = queryWithFilter({
+export const queryFiltered = query({
   args: {},
   handler: async (ctx) => {
-    return await ctx.db
-      .query("presence")
-      .filterWith(async (presence) => {
+    return await filter(ctx.db
+      .query("presence"), async (presence) => {
         return presence.updated > 10;
       })
       .first();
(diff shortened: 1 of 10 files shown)


convex/customFunctionExample.ts
```diff
@@ -13,6 +13,7 @@ import { getUserByTokenIdentifier } from "./lib/withUser";
 import { Rules } from "convex-helpers/server/rowLevelSecurity";
 import { DataModel, Doc } from "./_generated/dataModel";
 import { SessionIdArg, vSessionId } from "convex-helpers/server/sessions";
+import { ReaderWithFilter } from "convex-helpers/server/filterWith.js";
```
ldanilek, line 16: Why does this need the .js to work?

ianmacartney: You need to add an entrypoint to package.json for the non-node-routing to find it

packages/convex-helpers/server/filterWith.ts
```diff
@@ -0,0 +1,216 @@
...
+  }
+  filter(predicate: (q: FilterBuilder<T>) => Expression<boolean>): any {
+    return this.fullTableScan().filter(predicate);
+  }
+  filterWith(predicate: Predicate<T>): any {
+    return this.fullTableScan().filterWith(predicate);
+  }
+  order(order: "asc" | "desc"): OrderedQuery<T> {
+    return this.fullTableScan().order(order);
+  }
```
ianmacartney, line 164: this means if you call `.order` then you can no longer call `.filterWith`

packages/convex-helpers/server/filterWith.ts
```diff
@@ -0,0 +1,216 @@
...
+    return new QueryWithFilter(this.q.fullTableScan(), this.p);
+  }
+  withIndex<IndexName extends keyof Indexes<T>>(
+    indexName: IndexName,
+    indexRange?:
+      | ((
+          q: IndexRangeBuilder<DocumentByInfo<T>, NamedIndex<T, IndexName>, 0>
+        ) => IndexRange)
+      | undefined
+  ): Query<T> {
```
ianmacartney, line 142: this means you can't do `filterWith` if you use an index

packages/convex-helpers/server/filterWith.ts
```diff
@@ -0,0 +1,216 @@
...
+    return new QueryWithFilter(this.q.withIndex(indexName, indexRange), this.p);
+  }
+  withSearchIndex<IndexName extends keyof SearchIndexes<T>>(
+    indexName: IndexName,
+    searchFilter: (
+      q: SearchFilterBuilder<DocumentByInfo<T>, NamedSearchIndex<T, IndexName>>
+    ) => SearchFilter
+  ): OrderedQuery<T> {
```
ianmacartney, line 150: this means you can't do `filterWith` if you use a search index

packages/convex-helpers/server/filterWith.ts
```diff
@@ -0,0 +1,216 @@
...
+    return new QueryWithFilter(
+      this.q.withSearchIndex(indexName, searchFilter),
+      this.p
+    );
+  }
+  filter(predicate: (q: FilterBuilder<T>) => Expression<boolean>): any {
```
ianmacartney, line 156: this unfortunately results in the rest of the chain being `any`. unfortunately it also can't be `this` since `QueryWithFilter` doesn't extend it

packages/convex-helpers/server/filterWith.ts
```diff
@@ -0,0 +1,216 @@
...
+    return this.fullTableScan().filter(predicate);
+  }
+  filterWith(predicate: Predicate<T>): any {
```
ianmacartney, line 159: this can be a more specific type since it doesn't exist on the super type

packages/convex-helpers/server/filterWith.ts
```diff
@@ -0,0 +1,216 @@
...
+    return this.fullTableScan().filterWith(predicate);
+  }
+  order(order: "asc" | "desc"): OrderedQuery<T> {
```
ianmacartney, line 162: same here

packages/convex-helpers/server/filterWith.ts
```diff
@@ -0,0 +1,216 @@
...
+    return new QueryWithFilter(this.q.filter(predicate), this.p) as this;
+  }
+  filterWith(predicate: Predicate<T>): this {
+    return new QueryWithFilter(this.q, async (d) => {
+      return (await this.p(d)) && (await predicate(d));
+    }) as this;
+  }
+  order(order: "asc" | "desc"): QueryWithFilter<T> {
+    return new QueryWithFilter(this.q.order(order), this.p);
+  }
```
ianmacartney, line 64: this means you can call `.order` multiple times on a query

ianmacartney (changes requested):

packages/convex-helpers/server/filterWith.ts
ldanilek: can we make it be `this` and do a cast in the function?

packages/convex-helpers/server/filterWith.ts
ldanilek: yeah darn. you sort of can anyway in javascript because `QueryImpl` implements both Query and OrderedQuery, but it would be good for the types to not allow it

ianmacartney (approved):
So good! 
Can you drop a little section in the README.md about it for discoverability, linking to the Stack post? Can be copied from the docstring comment.

ianmacartney:
It'd also be nice to have `filterExample.ts` that exercise it a bit, possibly tied to a frontend component that lets you see some results / as manual testing until I set up more robust testing.

ldanilek:
> It'd also be nice to have `filterExample.ts` that exercise it a bit, possibly tied to a frontend component that lets you see some results / as manual testing until I set up more robust testing.

I was meaning to do this but got caught up today. Will try tomorrow

PR #73, opened 2024-03-07
SessionProvider: allow session IDs from a supplied function

tiernacity:
Hi, proposing a small quality-of-life improvement. Motivations:

- In my case: testing on iOS devices on my local network was failing, because crypto.randomUUID is not available when the protocol is not `https://` but `http://`. This change allows me to create session IDs with the CUID generator (which doesn't complain) instead
- Hypothetically: one might want or need a different format of session IDs, for some reason


ianmacartney (approved):
Very useful, thank you @tiernacity

PR #77, opened 2024-03-08
Make `npm run test` pass

sshader:
We have several `.test.ts` files that aren't actually vitest tests, one of which does not actually run (`zod.test`).

This excludes them. This also fixes the fake ConvexReactClient + mock test.


ianmacartney (approved):
thank you - The testing story for helpers is very weak rn, partly due to my lack of vitest experience. very open to ideas / contributions around it

PR #80, opened 2024-03-11
Add support for tests against a local backend

sshader:
Adds `ConvexTestingHelper` to the `convex-helpers` npm package + a simple example of running tests with a local backend.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -1,0 +1,32 @@
+set fallback := true
+set shell := ["bash", "-uc"]
+set windows-shell := ["sh", "-uc"]
+
+# `just --list` (or just `just`) will print all the recipes in
+# the current Justfile. `just RECIPE` will run the macro/job.
+#
+# In several places there are recipes for running common scripts or commands.
+# Instead of `Makefile`s, Convex uses Justfiles, which are similar, but avoid
+# several footguns associated with Makefiles, since using make as a macro runner
+# can sometimes conflict with Makefiles desire to have some rudimentary
+# understanding of build artifacts and associated dependencies.
+#
+# Read up on just here: https://github.com/casey/just
+
+_default:
+  @just --list
+
+set positional-arguments
+
+reset-local-backend:
+  cd $CONVEX_LOCAL_BACKEND_PATH; rm -rf convex_local_storage && rm -f convex_local_backend.sqlite3
+
+# (*) Run the open source convex backend on port 8000
+run-local-backend *ARGS:
+  cd $CONVEX_LOCAL_BACKEND_PATH && just run-local-backend
+
+# Taken from https://github.com/get-convex/convex-backend/blob/main/Justfile
+# (*) Run convex CLI commands like `convex dev` against local backend from `just run-local-backend`.
+convex *ARGS:
+  npx convex "$@" --admin-key 0135d8598650f8f5cb0f30c34ec2e2bb62793bc28717c8eb6fb577996d50be5f4281b59181095065c5d0f86a2c31ddbe9b597ec62b47ded69782cd --url "http://127.0.0.1:8000"
+
--- a/Justfile
+++ b/Justfile
@@ -21,14 +21,14 @@ set positional-arguments
 reset-local-backend:
   cd $CONVEX_LOCAL_BACKEND_PATH; rm -rf convex_local_storage && rm -f convex_local_backend.sqlite3
 
-# (*) Run the open source convex backend on port 8000
+# (*) Run the open source convex backend on port 3210
 run-local-backend *ARGS:
-  cd $CONVEX_LOCAL_BACKEND_PATH && just run-local-backend
+  cd $CONVEX_LOCAL_BACKEND_PATH && just run-local-backend --port 3210
 
 # Taken from https://github.com/get-convex/convex-backend/blob/main/Justfile
 # (*) Run convex CLI commands like `convex dev` against local backend from `just run-local-backend`.
 # This uses the default admin key for local backends, which is safe as long as the backend is
 # running locally.
 convex *ARGS:
-  npx convex "$@" --admin-key 0135d8598650f8f5cb0f30c34ec2e2bb62793bc28717c8eb6fb577996d50be5f4281b59181095065c5d0f86a2c31ddbe9b597ec62b47ded69782cd --url "http://127.0.0.1:8000"
+  npx convex "$@" --admin-key 0135d8598650f8f5cb0f30c34ec2e2bb62793bc28717c8eb6fb577996d50be5f4281b59181095065c5d0f86a2c31ddbe9b597ec62b47ded69782cd --url "http://127.0.0.1:3210"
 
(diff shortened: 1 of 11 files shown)


Justfile
```diff
@@ -0,0 +1,32 @@
...
+  cd $CONVEX_LOCAL_BACKEND_PATH; rm -rf convex_local_storage && rm -f convex_local_backend.sqlite3
+
+# (*) Run the open source convex backend on port 8000
+run-local-backend *ARGS:
+  cd $CONVEX_LOCAL_BACKEND_PATH && just run-local-backend
+
+# Taken from https://github.com/get-convex/convex-backend/blob/main/Justfile
+# (*) Run convex CLI commands like `convex dev` against local backend from `just run-local-backend`.
+convex *ARGS:
+  npx convex "$@" --admin-key 0135d8598650f8f5cb0f30c34ec2e2bb62793bc28717c8eb6fb577996d50be5f4281b59181095065c5d0f86a2c31ddbe9b597ec62b47ded69782cd --url "http://127.0.0.1:8000"
```
ianmacartney, line 31: what's the deal with this admin key?

README.md
```diff
@@ -169,3 +169,20 @@ field, and `brandedString`. To learn more about branded strings see
 runtime values. (This is controvercial and not required to use the above).
 
 See more [in the convex-helpers README](./packages/convex-helpers/README.md).
+
+## Testing with a local backend
+[`convex/example.test.ts`](./convex/example.test.ts) demonstrates testing Convex functions by running them against a local backend.
+
```
ianmacartney, line 175:
```suggestion
See [this Stack article](https://stack.convex.dev/testing-with-local-oss-backend) for more information.

```

packages/convex-helpers/testing.ts
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  withIdentity(
+    identity: Omit<UserIdentity, "tokenIdentifier">
+  ): Pick<ConvexClient, "mutation" | "action" | "query"> {
+    return {
+      mutation: (functionReference, args) => {
+        (this.client as any).setAdminAuth(this._adminKey, identity);
+        return this.client.mutation(functionReference, args).finally(() => {
+          this.client.client.clearAuth();
```
ianmacartney, line 60: is there interleaving edge case behavior if you're running multiple mutations in parallel during a test? Or will the synchronous setAdmin, then start the call, mostly be fine? Worst case you clear someone else's auth or something

package-lock.json
```diff
@@ -118,29 +111,27 @@
     },
     "node_modules/@babel/compat-data": {
       "version": "7.23.5",
-      "resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.23.5.tgz",
-      "integrity": "sha512-uU27kfDRlhfKl+w1U6vp16IuvSLtjAxdArVXPa9BvLkrr7CYIsxH5adpHObeAGY/41+syctUWOZ140a2Rvkgjw==",
       "dev": true,
+      "license": "MIT",
```
ianmacartney, line 115: this seems to happen a lot - one command will add a license and lose the resolved / integrity, and npm update will put those back. Do you know why? Or what command did you run that did this?

package-lock.json
```diff
@@ -504,723 +467,317 @@
...
-      "integrity": "sha512-ga8y9v9uyeiLdpKddhxYQkxNDrfvuPrlFb0N1qnZZByvcElJaXthF1UhvCh9TLWJBEHeNtdnbysW7Y6Uq8CVng==",
-      "dev": true
+      "dev": true,
+      "license": "MIT"
     },
     "node_modules/@types/react": {
-      "version": "18.2.64",
-      "resolved": "https://registry.npmjs.org/@types/react/-/react-18.2.64.tgz",
-      "integrity": "sha512-MlmPvHgjj2p3vZaxbQgFUQFvD8QiZwACfGqEdDSWou5yISWxDQ4/74nCAwsUiX7UFLKZz3BbVSPj+YxeoGGCfg==",
+      "version": "18.2.48",
```
ianmacartney, line 737: we seem to be going backwards on a few things - did you just resolve conflict by choosing your package-lock? I had just updated a bunch of dependencies. Could you reset the package-lock and re-install jest with the hopes of a smaller, more focused change just adding jest?

package-lock.json
```diff
@@ -1691,32 +1307,13 @@
...
-        },
-        "react": {
-          "optional": true
-        },
-        "zod": {
-          "optional": true
-        }
-      }
+      "resolved": "packages/convex-helpers",
+      "link": true
```
ianmacartney, line 1311: this change in particular gets annoying - you can restore it by doing `npm i ./packages/convex-helpers` and deleting the convex-helpers package at the bottom of the lockfile

package-lock.json
```diff
@@ -4438,11 +4018,35 @@
...
+        "hono": {
+          "optional": true
+        },
+        "react": {
+          "optional": true
+        },
+        "zod": {
+          "optional": true
+        }
+      }
```
ianmacartney, line 4049: delete this one if it gets added - it's redundant when you install via path, which the publish script does

ianmacartney (approved):
good pending lockfile changes and a few small notes

package-lock.json
sshader: I ran `npm install`. I get a fairly large diff when I do this on main too
```
rm -rf node_modules
rm -rf package-lock.json
npm install
```

I'm a little tempted to just check that in as a separate change to try and get some more reasonable diffs here

packages/convex-helpers/testing.ts
sshader: Yeah I actually want to switch this to using the HTTP client so it's less stateful, but the change to add the second argument to `setAdminAuth` requires an npm release (but I landed the change already)

package-lock.json
sshader: Ok yeah I split out the first commit which is just doing those steps off of main (which is a pretty large change, so I suspect we're not keeping our package-lock.json up to date properly when updating things). And I re-did these steps when adding the @types/jest dependency.

ianmacartney: When I do `rm -rf node_modules/ package-lock.json ; npm i; npm i ./packages/convex-helpers` I get a different result. I'll push those result to main. not sure if there's a better way to keep the relative package up to date. I had trouble getting it to update in the package lock without explicitly re-importing it after updating versions

convex/testingFunctions.ts
```diff
@@ -0,0 +1,56 @@
...
+  args: {},
+  input: async (_ctx, _args) => {
+    if (process.env.IS_TEST === undefined) {
+      throw new Error(
+        "Calling a test only function in an unexpected environment"
+      );
+    }
+    return { ctx: {}, args: {} };
+  },
+});
```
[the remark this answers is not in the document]
sshader: FWIW it'll vary from developer to developer whether they're using `customCtx` or `verify` so I kind of like the most explicit thing so there's no potential confusion around what to return.

Like if I was already using a `customQuery` handling a sessionId argument, I think it would be more obvious how to add this `IS_TEST` check to make a test only version of my query wrapper than if I used `customCtx` / `verify`

ianmacartney:
makes sense

PR #81, opened 2024-03-12
json helper

ianmacartney:
Add a helper for returning json from http endpoints.

### DIFF
--- a/packages/convex-helpers/server/index.ts
+++ b/packages/convex-helpers/server/index.ts
@@ -203,3 +203,39 @@ export function crud<
     >,
   };
 }
+
+export type JSONValue =
+  | null
+  | boolean
+  | undefined
+  | number
+  | string
+  | JSONValue[]
+  | { [key: string]: JSONValue };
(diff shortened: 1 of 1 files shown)


packages/convex-helpers/server/index.ts
```diff
@@ -203,3 +203,39 @@ export function crud<
...
+ *     return json({ message: "Hello, world!" });
+ *   }),
+ * });
+ * ```
+ *
+ * @param data The data you want to return as JSON.
+ * @returns A Response object with the data as JSON.
+ */
+export function json(data: JSONValue) {
+  return new Response(JSON.stringify(data), {
```
sshader, line 236: https://developer.mozilla.org/en-US/docs/Web/API/Response/json_static exists (and is implemented). But I think the types only exist in newer versions of TypeScript.

Would prefer pointing to that instead of creating our own version

ianmacartney:
wow thanks, I didn't know about that one.

PR #89, opened 2024-03-22
Dual package

thomasballinger:
Compile ESM and CJS builds and types and use exports to point to them.

This resolves the warnings [here](https://arethetypeswrong.github.io/?p=convex-helpers%400.1.27) and makes ts-jest work in https://github.com/get-convex/convex-chess/pull/19/files without adding the exception not to compile these files.

I'm not sure this needs to be a dual package though. If we could get away with ESM-only the build would be simpler. That's the way we should recommend developers build things in Convex.

### DIFF
--- a/packages/convex-helpers/generate-exports.mjs
+++ b/packages/convex-helpers/generate-exports.mjs
@@ -1,0 +1,93 @@
+import fs from "node:fs";
+import path from "node:path";
+import process from "node:process";
+import { fileURLToPath } from "node:url";
+
+const __dirname = path.dirname(fileURLToPath(new URL(import.meta.url)));
+
+function directoryContents(dirname) {
+  return fs
+    .readdirSync(path.join(__dirname, dirname))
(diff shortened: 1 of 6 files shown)


packages/convex-helpers/generate-exports.mjs
```diff
@@ -0,0 +1,93 @@
...
+    console.error(
+      "`package.json` exports are not correct. Copy exports from above or run"
+    );
+    console.error("node generate-exports.mjs | pbcopy");
+    console.error("and paste into package.json.");
+    process.exit(1);
+  }
+}
+
+checkPackageJsonExports();
```
ianmacartney, line 93: is this supposed to also write the exports? I only see the verification

thomasballinger: yeah just verification, to avoid the checked in package.json differing from what's used, e.g. accidentally publishing changes not checked in to the repo

ianmacartney:
Closing, since esm seems fine for us for now

PR #90, opened 2024-03-23
Simple compile step in packaging

thomasballinger:
Compile TypeScript to ESM .js, .d.ts, and .d.ts.map files.

This resolves the issue in https://github.com/get-convex/convex-helpers/pull/80 that required adding node_modules/convex-helpers to the list of files to transpile.

This may be the setup to recommend to others: no CJS so no dual packages, simple tsconfig.json setup, compile step but no bundling, and perhaps this fancy "development" setup although that one could use more testing. If configured right it should allow skipping the compile step when this package is installed in another project from this repo checkout out locally.


packages/convex-helpers/package.json
```diff
@@ -30,11 +79,12 @@
...
   ],
   "scripts": {
-    "build": "tsc",
+    "build": "node generate-exports.mjs && tsc",
     "prebuild": "npm run test",
     "test": "tsc --project tsconfig.test.json",
     "clean": "rm -rf dist",
-    "watch": "chokidar '*.ts' 'server/*.ts' 'react/*.ts' 'tsconfig.json' 'package.json' -c 'npm run build' --initial"
+    "watch": "chokidar '*.ts' 'server/*.ts' 'react/*.ts' 'tsconfig*.json' 'package.json' -c 'npm run build' --initial",
+    "arethetypeswrong": "npx attw $(npm pack)"
```
ianmacartney, line 87: presumably this should be called by the publish script?

packages/convex-helpers/tsconfig.json
```diff
@@ -112,6 +112,8 @@
   "exclude": [
     "node_modules",
     "dist",
-    "**/*.test.ts"
+    "**/*.test.ts",
+    "*.test.ts",
+    "zod/*.test.ts"
```
ianmacartney, line 117: why is this one a special case?

ianmacartney (approved):
I mean, if we can get away with this one, it seems way nicer. Does it solve the cjs/mjs issue with chess? I wasn't sure

thomasballinger:
The helpers we need are exposed in https://github.com/get-convex/convex-backend/commit/698b96a8ef4dc4961c717131ae1cb06540f6838f, no release yet

packages/convex-helpers/tsconfig.json
thomasballinger: shouldn't be here, was troubleshooting

packages/convex-helpers/package.json
thomasballinger: it's a lint for troubleshooting, we could run it with tests

ianmacartney:
Also npx shouldn’t be necessary in scripts- they add the node modules bin
to the path iirc

packages/convex-helpers/tsconfig.json
ianmacartney, line 117:
```suggestion
    "**/*.test.ts",
    "*.test.ts"
```

ianmacartney (commented):
Is this ready to go now that the types are exposed from convex @thomasballinger ?

package-lock.json
```diff
@@ -1862,26 +2251,8 @@
...
-        },
-        "react": {
-          "optional": true
-        },
-        "zod": {
-          "optional": true
-        }
-      }
+      "resolved": "packages/convex-helpers",
+      "link": true
```
ianmacartney, line 2255: so this change keeps getting recommended by renovate but when I did the install like `npm i ./packages/convex-helpers` it un-does it, and doing `npm i` this way I think was not doing what I expected or I had trouble doing npm link or something. Do you know what this is about @thomasballinger - is this the "right" way to do it, vs. the `file:packages/convex-helpers` mode it's replacing? 
It would be nice to do it the way renovate keeps overriding, but if it's always fighting with the publish script it'll be annoying.

thomasballinger: Running `npm i ./package/convex-helpers` doesn't undo this for me and I still see live updating — so it seems to work.

I think we can stop using `npm link` and just rely on this "development" thing. I removed all npm links on my machine and still see these updating live. I'll remove the npm link stuff in this PR.

thomasballinger:
This works for everything I try to do with it, but that's just loading files in the top-level project that import from convex-helpers. Is there anything else to check here?

package-lock.json
thomasballinger: Interestingly `"convex-helpers": "file:packages/convex-helpers"` always seems to result in a symlink at `node_modules/convex-helpers`, no `npm link` needed

thomasballinger: Oh what version of npm are you using? that could be the different behavior we see

thomasballinger:
Asking online, it sounds like the "development" export condition is not a good idea because some bundlers will be looking for a development build but not be able to transpile the TypeScript. I wonder if we should use a different condition or strip the development condition.

I think we don't need the npm link but we do need the watch process, just like you mentioned in person for the `convex dev` bundling process (which you were also right about, esbuild does not by default use the "development" export condition)

thomasballinger:
(from https://x.com/AndaristRake/status/1776000460744237333)

packages/convex-helpers/server/index.ts
```diff
@@ -12,7 +12,7 @@ import {
   PaginationResult,
 } from "convex/server";
 import { GenericId, Infer, ObjectType, Validator, v } from "convex/values";
-import { Expand } from "../index.js";
+import { Expand } from "..";
```
ianmacartney, line 15: I don't think this was intentional?

thomasballinger: thanks, was reverting some formatting changes

thomasballinger:
I'll land this without the publish script changes you made and you can add them? should be ready now

thomasballinger:
Oh whoops I force-pushed your publish script changes away, I'll add them back

PR #110, opened 2024-04-15
Sessions with SSR

ianmacartney:
Support SSR with the SessionProvider by supplying a default value on the first pass client and server-side, and skipping queries that use that value. opt-in via `ssrFriendly`

### DIFF
--- a/packages/convex-helpers/react/sessions.ts
+++ b/packages/convex-helpers/react/sessions.ts
@@ -74,6 +74,11 @@ export const SSR_DEFAULT = "SSR default session ID" as SessionId;
  *  - localStorage is shared between tabs, but not browser profiles.
  * @param storageKey - Key under which to store the session ID in the store
  * @param idGenerator - Function to return a new, unique session ID string. Defaults to crypto.randomUUID
+ * @param ssrFriendly - Returns SSR_DEFAULT on the first pass, so hydration
+ *   doesn't mismatch and an ID isn't generated server-side.
+ *   useSessionQuery will skip queries until there is a real value.
+ *   However, if you're using useSessionId, consider checking for SSR_DEFAULT.
+ *   Defaults to false, where it will always return a valid id.
  * @returns A provider to wrap your React nodes which provides the session ID.
  * To be used with useSessionQuery and useSessionMutation.
  */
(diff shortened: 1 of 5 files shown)


packages/convex-helpers/react/sessions.ts
```diff
@@ -66,25 +74,46 @@ type SessionArgsArray<Fn extends SessionFunction<"mutation" | "action">> =
...
   // Get or set the ID from our desired storage location.
   const useStorageOrDefault = useStorage ?? useSessionStorage;
-  const [sessionId, setSessionId] = useStorageOrDefault(storeKey, initialId);
+  const [sessionId, setSessionId] = useStorageOrDefault(
+    storeKey,
+    ssrFriendly ? SSR_DEFAULT : idGen()
+  );
+
+  const [initial, setInitial] = useState(true);
+  if (ssrFriendly) {
```
ianmacartney, line 105: against the rules but should be constant

xixixao: You can also put the if condition inside the useEffect

packages/convex-helpers/react/sessions.ts
```diff
@@ -93,21 +122,24 @@ export const SessionProvider: React.FC<{
...
-  return React.createElement(
-    SessionContext.Provider,
-    { value: { sessionId, refreshSessionId } },
-    children
+  const value = useMemo(
+    () => ({
+      sessionId: initial && ssrFriendly ? SSR_DEFAULT : sessionId,
+      refreshSessionId,
+    }),
+    [initial, ssrFriendly, sessionId, refreshSessionId]
```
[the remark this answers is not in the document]
xixixao: It is not, as long as you document that the object should always be destructured at callsite. The changing value has no impact on how many renders are triggered (changing React Provider value doesn't cause a rerender).

packages/convex-helpers/react/sessions.ts
```diff
@@ -66,25 +74,46 @@ type SessionArgsArray<Fn extends SessionFunction<"mutation" | "action">> =
  *  - localStorage is shared between tabs, but not browser profiles.
```
xixixao, line 74: Would be nice to note that sessionStorage is the default

PR #114, opened 2024-04-16
Sessions

ianmacartney:
Store the session ID as undefined by default, rather than the weird string.

Previously if your app didn't successfully execute the useEffect you'd end up with the SSR_DEFAULT in your localstorage and just keep using that later. 

Also some feedback from @xixixao

### DIFF
--- a/packages/convex-helpers/react/sessions.ts
+++ b/packages/convex-helpers/react/sessions.ts
@@ -101,10 +101,12 @@ export const SessionProvider: React.FC<{
     ssrFriendly ? undefined : idGen()
   );
 
+  const [initial, setInitial] = useState(true);
   // Generate a new session ID on first load.
   // This is to get around SSR issues with localStorage.
   useEffect(() => {
     if (!sessionId || sessionId === SSR_DEFAULT) setSessionId(idGen());
+    if (ssrFriendly && initial) setInitial(false);
   }, [setSessionId, sessionId]);
 
   const refreshSessionId = useCallback<RefreshSessionFn>(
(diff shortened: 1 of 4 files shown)


packages/convex-helpers/react/sessions.ts
```diff
@@ -98,18 +98,16 @@ export const SessionProvider: React.FC<{
...
-    // This is to get around SSR issues with localStorage.
-    useEffect(() => {
-      if (sessionId === SSR_DEFAULT) setSessionId(idGen());
-      if (initial) setInitial(false);
-    }, [setSessionId, sessionId, setInitial, initial]);
-  }
+  // Generate a new session ID on first load.
+  // This is to get around SSR issues with localStorage.
+  useEffect(() => {
+    if (!sessionId || sessionId === SSR_DEFAULT) setSessionId(idGen());
```
xixixao, line 108: You're keeping the SSR_DEFAULT for backwards compatibility?

packages/convex-helpers/react/sessions.ts
```diff
@@ -124,10 +122,11 @@ export const SessionProvider: React.FC<{
   );
   const value = useMemo(
     () => ({
-      sessionId: initial && ssrFriendly ? SSR_DEFAULT : sessionId,
+      sessionId:
+        ssrFriendly && initial ? SSR_DEFAULT : sessionId || SSR_DEFAULT,
```
xixixao, line 126: Maybe
```suggestion
        ssrFriendly && initial ? SSR_DEFAULT : sessionId ?? SSR_DEFAULT,
```

xixixao (approved):

packages/convex-helpers/react/sessions.ts
ianmacartney: yeah or as a backstop

packages/convex-helpers/react/sessions.ts
ianmacartney: Yeah I thought about that too. I'm thinking an empty string is also a bad value, if they had manually edited their localStorage or something. If they have a custom session ID generator maybe they want the empty string sentinel value, but I'll just wait for someone to complain on that I think

PR #136, opened 2024-06-12
Query Cache

jamwt:
Implementation of a query cache that can dramatically speed up re-rendering of values as SPA routes change and components are unmounted and remounted.

----


ianmacartney:
A couple thoughts just coming to me:
1. We could move this into the npm package, maybe namespaced under a `react/cache` entrypoint in a single file.
2. I wonder if we should change the tanstack-style `useQuery` hook in helpers to also do this, but only if it's in the cache provider - and maybe push folks to use that one. As it stands there will be 3 options for `useQuery`: vanilla, this one, and tanstack-style. there's a usecase for vanilla - if you don't want to use the caching - but that implies there should be an opt-out for the tanstack-style one too. So either 2 versions (TanStack to also get caching, vanilla if you want to opt out), or 4 versions (full permutations - or an extra arg to tanstack about opting out). I kinda wish I had gone with a single object arg for the tanstack one after all (`{ queryFn, queryArgs }`) 

Was it intentional to keep it out of the npm package for now, just to point people to it?

jamwt:
1. I just don't know how to do that, but otherwise -- yeah, would love to have it in the package.
2. Yeah, I debated the tanstack-style vs. this. Any ideas about how to support both without replicating a ton of code?

jamwt:
Okay, I

1. Refactored everything to be centered around `useQueries` so that the tanstack-style wrapper can use this cache.
2. Fixed error handling
3. Moved this into the `convex-helpers` package, in the react module.

README.md
```diff
@@ -113,6 +114,25 @@ Related files:
...
+  a configurable cache provider you put in your react app's root.
+- [useQuery.tsx](./src/hooks/cache/useQuery.tsx) is a drop-in replacement for the
+  `convex/react` equivalent that utilizes this configured cache.
+- [CacheExample.tsx](./src/components/CacheExample.tsx) is a demo component demonstrating
+  the cache's `useQuery` to reuse app-side values when child components are
+  unmounted and re-mounted.
+
+To use the query cache, copy the entire `src/hooks/cache` directory into your project. The
+`cache/index.ts` re-exports the things you'll need.
+
```
ianmacartney, line 135: let's move this to the README in packages/convex-helpers, and update the paths

README.md
```diff
@@ -29,6 +29,7 @@ find the npm package in [./packages/convex-helpers](./packages/convex-helpers).
 | [Presence](#presence)
 | [Throttling via single-flighting](#throttling-client-side-requests-by-single-flighting)
 | [Stable query results via useStableQuery](#stable-query-results-via-usestablequery)
+| [Query caching with ConvexQueryCacheProvider](#query-caching)
```
ianmacartney, line 32: let's move this up to the list above, linking to the readme entry there

packages/convex-helpers/package.json
```diff
@@ -24,6 +24,18 @@
...
+      "default": "./dist/react/cache/core.js"
+    },
+    "./react/cache/hooks": {
+      "types": "./dist/react/cache/hooks.d.ts",
+      "default": "./dist/react/cache/hooks.js"
+    },
+    "./react/cache/provider": {
+      "types": "./dist/react/cache/provider.d.ts",
+      "default": "./dist/react/cache/provider.js"
+    },
```
ianmacartney, line 38: Is it a React requirement that these be 3 separate files? I know the organization is nice, but I'd prefer to have it all in `react/cache.ts` so it's one entrypoint per feature

packages/convex-helpers/tsconfig.json
```diff
@@ -4,16 +4,16 @@
...
-    "composite": true,                                /* Enable constraints that allow a TypeScript project to be used with project references. */
+    "composite": true /* Enable constraints that allow a TypeScript project to be used with project references. */,
     // "tsBuildInfoFile": "./.tsbuildinfo",              /* Specify the path to .tsbuildinfo incremental compilation file. */
     // "disableSourceOfProjectReferenceRedirect": true,  /* Disable preferring source files instead of declaration files when referencing composite projects. */
     // "disableSolutionSearching": true,                 /* Opt a project out of multi-project reference checking when editing. */
     // "disableReferencedProjectLoad": true,             /* Reduce the number of projects loaded automatically by TypeScript. */
 
     /* Language and Environment */
-    "target": "ESNext",                                  /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */
+    "target": "ESNext" /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */,
```
ianmacartney, line 14: I think these are unintentional - maybe our prettier config now fights with the default tsconfig formatting?

packages/convex-helpers/package.json
jamwt: Yeah. SSR messes with this. "use client" (hook) code and provider code can't be in the same file.

packages/convex-helpers/tsconfig.json
jamwt: I dunno. I can revert... is that what you're suggesting?

README.md
jamwt: Good point, will do.

packages/convex-helpers/package.json
ianmacartney: ... but the hooks file doesn't have "use client" at the top? At the least I'm taking out core, since none of the code is shared between them

PR #137, opened 2024-06-12
Manual pagination helper

ldanilek:
Create a new convex helper: manual pagination!

See the README for usages.
also the tests.

----


ianmacartney (commented):
ah shoot never hit submit here, sorry

packages/convex-helpers/README.md
```diff
@@ -88,10 +88,10 @@ const posts = await asyncMap(
       "postCategories",
       "categoryId",
       "postId",
-      post._id,
+      post._id
```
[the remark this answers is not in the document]
ldanilek:
```
$ npx prettier --version
3.2.5
```

packages/convex-helpers/README.md
```diff
@@ -623,10 +623,266 @@ export const lastCountLongerThanName = query({
...
   },
 });
 ```
+
+## Manual Pagination
+
+### Benefits of manual pagination
+
+- You can preprocess the table to store all page boundaries, allowing skipping to an
+  arbitrary page.
```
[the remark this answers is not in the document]
ldanilek: yeah i was thinking it could be a component. i was thinking you could do it with function references, where the passed-in function takes in a start index key, calls `getPage`, and returns the end index key. but the write-time update hook is more complex

packages/convex-helpers/server/pagination.ts
```diff
@@ -0,0 +1,293 @@
...
+      .order(order);
+    for await (const doc of query) {
+      if (page.length >= absoluteLimit) {
+        hasMore = true;
+        break;
+      }
+      page.push(doc);
+      indexKeys.push(getIndexKey(doc, indexFields));
+    }
+  }
```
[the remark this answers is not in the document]
ldanilek: whoops good call.
i accidentally thought the `break;` would break out of the outer loop. to prevent such confusion, i'll change this to an early return.

packages/convex-helpers/README.md
```diff
@@ -623,10 +623,266 @@ export const lastCountLongerThanName = query({
...
-      (c) => c.counter > c.name.length,
+      (c) => c.counter > c.name.length
     )
       .order("desc")
       .first();
   },
 });
 ```
+
+## Manual Pagination
```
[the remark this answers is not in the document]
ldanilek: ok i shortened it to just a few sentences and examples

packages/convex-helpers/README.md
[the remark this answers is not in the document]
ldanilek: okay i don't know what's going on with my prettier but it fixed itself automatically

packages/convex-helpers/README.md
```diff
@@ -623,10 +623,266 @@ export const lastCountLongerThanName = query({
...
+  page: page2,
+  indexKeys: indexKeys2,
+  hasMore,
+} = await getPage({
+  table: "messages",
+  startIndexKey: indexKeys[indexKeys.length - 1],
+});
+```
+
+Refresh the first page, with the same endpoints, so it stays adjacent to the
```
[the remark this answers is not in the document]
ldanilek: we can put this in the article or make a hook that does it. i left a single example

packages/convex-helpers/README.md
```diff
@@ -630,3 +630,74 @@ export const lastCountLongerThanName = query({
...
+});
+```
+
+Fetch the next page:
+
+```js
+const {
+  page: page2,
+  indexKeys: indexKeys2,
+  hasMore,
```
ianmacartney, line 663: I'm editing this to be `hasMore: hasMore2`

packages/convex-helpers/README.md
```diff
@@ -630,3 +630,74 @@ export const lastCountLongerThanName = query({
...
+```
+
+Fetch of a page between two fixed places in the index, allowing you to display
+continuous pages even as documents change and the queries reactively update:
+
+```js
+const { page } = await getPage(ctx, {
+  table: "messages",
+  startIndexKey,
+  endIndexKey,
```
ianmacartney, line 689: Is this how you'd do it? I have to wrap my head a little around the array of index keys. I'm pretty sure the first one is the absolute lower bound, and the last one is the absolute upper bound, but only after learning how it works
```suggestion
  startIndexKey: indexKeys[0],
  endIndexKey: indexKeys2[0],
```

packages/convex-helpers/README.md
```diff
@@ -630,3 +630,74 @@ export const lastCountLongerThanName = query({
...
 });
 ```
+
+## Manual Pagination
+
+Note Convex provides built-in pagination through `.paginate()` and
+`usePaginatedQuery()`.
+
+The `getPage` helper gives you more control of the pagination. You can specify
+the index ranges or do multiple paginations in the same query.
```
ianmacartney, line 640:
```suggestion
the index ranges or do multiple paginations in the same query.
An index range is all of the documents between two index keys.
An index key is an array of values for the fields in the specified index.
For example, for an index defined like `defineTable({ a: v.number(), b: v.string() }).index("my_index", ["a", "b"])`
an index key might be `[ 3 ]` or `[ 3, "abc" ]`. By default the index is the built-in "by_creation_time" index.
```

packages/convex-helpers/server/pagination.ts
```diff
@@ -0,0 +1,301 @@
...
+export type IndexKey = Value[];
+
+export type PageRequest<
+  DataModel extends GenericDataModel,
+  T extends TableNamesInDataModel<DataModel>,
+> = {
+  /// Request a page of documents from this table.
+  table: T;
+  /// Where the page starts. Default or empty array is the start of the table.
+  startIndexKey?: IndexKey;
```
ianmacartney, line 21: a future types rabbit hole to go down would be to type this based on a prefix of the `index` from `schema` specified or defaulting to a prefix of `[ number, GenericId<T> ]` unless they manually specify `indexFields`

packages/convex-helpers/server/pagination.ts
```diff
@@ -0,0 +1,301 @@
...
+  TableNamesInDataModel,
+} from "convex/server";
+
+export type IndexKey = Value[];
+
+export type PageRequest<
+  DataModel extends GenericDataModel,
+  T extends TableNamesInDataModel<DataModel>,
+> = {
+  /// Request a page of documents from this table.
```
ianmacartney, line 18: I know this is annoying, but `///` doesn't automatically generate docs on hover, at least for me. These need to be `/** Request a page of documents from this table. */` for single line, or `/** one line` ` * next line` ` */` style. VSCode should auto-generate some jsdoc syntax for functions. The nice thing is that then you get docs not only for functions but also for fields like this.

I've updated them this time - will push the commit

packages/convex-helpers/server/pagination.ts
```diff
@@ -0,0 +1,301 @@
...
+  table: T;
+  /// Where the page starts. Default or empty array is the start of the table.
+  startIndexKey?: IndexKey;
+  /// Whether the startIndexKey is inclusive. Default is false.
```
ianmacartney, line 22: Is there a standard here? I would have expected the flip (inclusive start, exclusive end).

ianmacartney (approved):

packages/convex-helpers/server/pagination.ts
ldanilek: when you do it like this, you can combine and refetch pages easily, using the index keys of the returned documents. 

Like if i fetch the first page and get results 1 to 10, i can refetch it by quering the range (0, 10] and i can fetch the next page as (10, infinity]. If you make the end exclusive, you need to get an index key >10 to refetch the first page, which you don't have if you only have documents 1 to 10.

packages/convex-helpers/README.md
ldanilek: When refetching you don't change `startIndexKey` -- it's the same as when you originally fetched. And you set `endIndexKey: indexKeys[indexKeys.length-1]` 

This works because the start is exclusive and the end is inclusive. This allows refetching without needing to fetch the subsequent page (you don't need `indexKeys2`)

packages/convex-helpers/README.md
ldanilek: Sg. We should also mention that index keys returned by getPage are always unique, because they end with the `_id` field

packages/convex-helpers/server/pagination.ts
ianmacartney: ah yes, that makes a lot of sense, thanks

packages/convex-helpers/README.md
ianmacartney: I'll leave it as is. I thought it was specifically targeting the "reconnect to avoid gaps" usecase but it's more general and I don't want to confuse things.

PR #166, opened 2024-07-11
Stricter tsconfig

ianmacartney:
Improving the tsconfig strictness so we don't cause type issues for developers importing this code

----

### DIFF
--- a/.vscode/convex.code-snippets
+++ b/.vscode/convex.code-snippets
@@ -140,7 +140,31 @@
 		"scope": "javascript,typescript",
 	},
 
-	"Covnex Test": {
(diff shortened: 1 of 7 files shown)


ianmacartney:
@ldanilek can you sanity check that the pagination assertions are ok? Should we add a check at the beginning like `indexKey.length >= startBound?.length ?? 0`?

ldanilek:
> @ldanilek can you sanity check that the pagination assertions are ok? Should we add a check at the beginning like `indexKey.length >= startBound?.length ?? 0`?

Sounds like a good assertion to add, yeah

ldanilek (approved):

PR #167, opened 2024-07-11
Fix type issues from npm 1.13

sshader:
Fixes three issues:
* `myCustomFunciton({ handler: (ctx, args: { foo: string } => ... })` now compiles
* `zodToConvex(z.object({ foo: z.number().optional() }))` now yields `{ foo?: number | undefined }` instead of `{ foo: number | undefined }`
* `literals("bar", "baz")` when used in an argument validator gets its TS type inferred as `"bar" | "baz"` instead of `string`. This fix required adding an `any` and throwing out some type information, so we're still looking into a better patch.

### DIFF
--- a/packages/convex-helpers/server/customFunctions.ts
+++ b/packages/convex-helpers/server/customFunctions.ts
@@ -487,6 +487,16 @@ type Expand<ObjectType extends Record<any, any>> =
       }
     : never;
 
+type ArgsForHandlerType<
+  OneOrZeroArgs extends [] | [Record<string, any>],
+  ModMadeArgs extends Record<string, any>,
+> =
+  ModMadeArgs extends Record<string, never>
+    ? OneOrZeroArgs
+    : OneOrZeroArgs extends [infer A]
+      ? [Expand<A & ModMadeArgs>]
+      : [ModMadeArgs];
+
 /**
  * A builder that customizes a Convex function, whether or not it validates
  * arguments. If the customization requires arguments, however, the resulting
@@ -501,8 +511,8 @@ export type CustomBuilder<
   Visibility extends FunctionVisibility,
 > = {
   <
-    ArgsValidator extends PropertyValidators | void,
-    ReturnsValidator extends GenericValidator | void,
+    ArgsValidator extends PropertyValidators | GenericValidator | void,
+    ReturnsValidator extends PropertyValidators | GenericValidator | void,
     ReturnValue extends ReturnValueForOptionalValidator<ReturnsValidator> = any,
     OneOrZeroArgs extends
       ArgsArrayForOptionalValidator<ArgsValidator> = DefaultArgsForOptionalValidator<ArgsValidator>,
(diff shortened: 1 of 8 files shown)


packages/convex-helpers/server/customFunctions.ts
```diff
@@ -501,8 +511,8 @@ export type CustomBuilder<
   Visibility extends FunctionVisibility,
 > = {
   <
-    ArgsValidator extends PropertyValidators | void,
-    ReturnsValidator extends GenericValidator | void,
+    ArgsValidator extends PropertyValidators | GenericValidator | void,
+    ReturnsValidator extends PropertyValidators | GenericValidator | void,
```
ianmacartney, line 515: For ArgsValidator, I don't think we're handling generic validator correctly yet? we're just spreading in ...fn.args when now it could be a validator object (or even a union of objects?) where adding extra args will need some extra handling in those cases

packages/convex-helpers/server/zod.ts
```diff
@@ -540,6 +540,18 @@ type ConvexUnionValidatorFromZod<T> = T extends z.ZodTypeAny[]
...
 
+type ConvexObjectValidatorFromZod<T extends Record<string, z.ZodTypeAny>> =
+  VObject<
+    ObjectType<{
+      [key in keyof T]: T[key] extends z.ZodTypeAny
+        ? ConvexValidatorFromZod<T[key]>
+        : never;
+    }>,
+    {
+      [key in keyof T]: ConvexValidatorFromZod<T[key]>;
```
ianmacartney, line 551: Why was the extends z.ZodTypeAny requiried in the above object type but not here?

packages/convex-helpers/server/zod.ts
```diff
@@ -637,22 +641,26 @@ type ConvexValidatorFromZod<Z extends z.ZodTypeAny> =
...
+                                                    undefined
+                                                  >,
+                                                  [
+                                                    Validator<T, "required", F>,
+                                                    VNull,
+                                                  ],
+                                                  "optional",
+                                                  ConvexValidatorFromZod<Inner>["fieldPaths"]
+                                                >
+                                              : never
```
ianmacartney, line 663: I wonder why prettier has changed - I often get reformatting on older files - maybe it pre-dated the prettierrc. 
Is there a better practice like pinning the prettier version too?

tests/convex/validators.ts
```diff
@@ -65,3 +66,12 @@ export const echo = internalQuery({
     return args;
   },
 });
+
+export const testLiterals = internalQuery({
```
ianmacartney, line 70: ty!

packages/convex-helpers/server/customFunctions.ts
sshader: Ooh yeah you're totally right. I'll probably punt on supporting `GenericValidator` for now and follow up with a `mergeObjects` helper

packages/convex-helpers/server/zod.ts
sshader: So tbh I kept pattern matching off of the `ConvexUnionValidatorFromZod` above until TS stopped complaining about a self referential type. Given the type bound on `T`, `T[key]` should always extend `z.ZodTypeAny`. I can do the same thing here too for consistency

ianmacartney (approved):
approved assuming this change looks good to you

ianmacartney:
cherry-picked

PR #174, opened 2024-07-12
Only publish the `dist` directory of the package

ianmacartney:
Instead of publishing packages/convex-helpers, publish packages/convex-helpers/dist
where we put only what we want to publish.

We can clean this up later to dynamically create the package.json / have it without the scripts in it. 

----


packages/convex-helpers/package.json
```diff
@@ -1,96 +1,104 @@
 {
   "name": "convex-helpers",
-  "version": "0.1.46",
+  "version": "0.1.47-alpha.2",
   "description": "A collection of useful code to complement the official convex package.",
   "type": "module",
   "exports": {
     ".": {
-      "types": "./dist/index.d.ts",
```
thomasballinger, line 8: it looks like you wouldn't be able to `npm install ../../convex-helpers` anymore with this since this package.json now points to files that don't exist

thomasballinger (commented):
It confuses me to see package.json semantics be wrong until the build happens, this means tools that read a package.json for module resolution instead of the published tarball won't work. I don't know all the implications of this so I'm not a fan, but exploring this space is really useful so curious to hear how this goes. If we want something that just works I'd do the stub directories for Node10 resolution.

thomasballinger:
oh sorry I missed the note of
> We can clean this up later to dynamically create the package.json / have it without the scripts in it.

that helps, that way we could have a normal package.json with working paths and construct a separate one with different paths. The other approach of doing the compile in-place, building the package, then deleting those files seems a little cleaner? because deleting files seems a little easier than modifying a package.json. We do some package.json modification of https://github.com/get-convex/convex-js/blob/main/package.json in https://github.com/get-convex/convex-js/blob/main/scripts/postpack.mjs and it's annoying.

packages/convex-helpers/package.json
ianmacartney: You can `npm install ../../convex-helpers/dist` now and it works a little more honestly

PR #178, opened 2024-07-17
Add folder/package.json entrypoints

ianmacartney:
Instead of shipping ts next to js, we can make entrypoints folders with a package.json that points to the dist directory. 
This was annoying to set up, but happy they only exist when code is being pushed.
If this doesn't solve [this issue on Discord](https://discord.com/channels/1019350475847499849/1252675728152072282/1263223018793078817) then it's probably not worth the breaking change?

----


thomasballinger (commented):
That's slick! If it works out I'm interested in copying it everywhere that needs it, curious to hear if it helps

ianmacartney:
Their issue wasn’t resolved with it but it’s here if anyone else finds
issues. Going to hold off since it does a bunch of things that aren’t
observable / hard to catch bugs before shipping / doesn’t npm pack as the
same

ianmacartney:
doesn't seem like it's been necessary

PR #186, opened 2024-07-26
Wildcard-export

ianmacartney:
Add wildcard exports to package to enable VSCode autocomplete for the library.
In addition to the explicit enumeration, so that we can still use arethetypeswrong.
involves moving *.index.ts files into *.ts files

----


thomasballinger (approved):
really cool

PR #197, opened 2024-08-05
add cors helper

ianmacartney:
a sketch of what it could look like w.r.t. composition vs. subclassing

----


ianmacartney:
```ts
const http = httpRouter();

const httpWithCORS = corsRouter(http, {
  allowedOrigins: ["*"], // allow regex
  allowCredentials: false, // if true, must specify non-* origins
  allowMethods: ["*"],
  allowHeaders: ["*"],
  exposeHeaders: [],
  maxBrowserCacheAge: 600,
});

// route with the one based on what you want.
httpWithCORS.route({
  allowMethods: // overrides
  // normal fields
  path: "/myfn",
  action: "POST",
  handler: myhandler,
});
```

packages/convex-helpers/server/cors.test.http.ts
```diff
@@ -0,0 +1,169 @@
...
+cors.route({
+  path: "/fact",
+  method: "DELETE",
+  handler: everythingHandler,
+});
+
+/**
+ * Non-CORS routes
+ */
+http.route({
```
sshader, line 52: Can you do things like route to `/foo` with method GET with CORS and `/foo` with method POST without CORS? From a quick scan, looks like yes. I guess the OPTIONS will only reflect the ones with the CORS router, which is fine, but also probably not a super common case

sshader (approved):

packages/convex-helpers/server/cors.test.http.ts
ianmacartney: Actually that's a bit of a rough edge: I believe if you route w/o CORS, then with CORS, the CORS router will include the non-cors method in the accepted methods

ianmacartney: I'll file an issue to track it

PR #202, opened 2024-08-08
[Function Metadata] Add helpers to generate an OpenAPI and JS specification

jordanhunt22:
Implements a `convex-helpers` CLI. Its initial functionality includes the ability to generate an OpenAPI and JS specification from a JSON that describes your function definitions.

Things to look at in this PR:
- confirm that we are comfortable with the testing coverage
- currently, there are two separate commands (`ts-api-spec`, `open-api-spec`). are we comfortable with this, are do we want to combine both into a single command like `generate-spec --format={format}`
- add any comments/suggestions for things i should include in the `readme` blurb about each command

Little things that still need to be done:
- we need to release Convex `1.15.0`, so I can get my changes for `npx convex function-spec` out
- then, we can require that our customers install `^1.15.0` to use convex helpers


cli/jsApiSpec.ts
```diff
@@ -0,0 +1,133 @@
+import fs from 'fs'
+import { execSync } from 'child_process';
+import { Command, Option } from "commander";
+import { ValidatorJSON } from 'convex/values'
+import { FunctionSpec } from './openApiSpec';
+import chalk from 'chalk';
+
+export const jsApiSpec = new Command("js-api-spec")
+    .summary("Generate a JavaScript API spec from a Convex function definition")
+    .argument("[filePath]", "The file name of the Convex function definition. If this argument is not provided, we will retrieve the function spec from your configured convex instance")
```
sshader, line 10:
```suggestion
    .argument("[filePath]", "The file name of the Convex function definition. If this argument is not provided, we will retrieve the function spec from your configured Convex deployment")
```

cli/jsApiSpec.ts
```diff
@@ -0,0 +1,133 @@
...
+        let content: string;
+        if (filePath && !fs.existsSync(filePath)) {
+            console.error(`File ${filePath} not found`);
+            process.exit(1);
+        }
+        if (filePath) {
+            content = fs.readFileSync(filePath, 'utf-8');
+        } else {
+            const flags = prod ? '--prod' : '';
+            const output = execSync(`npx convex function-spec ${flags}`);
```
sshader, line 28: might be nice to check that this executed successfully and print a nice error message if not?

cli/jsApiSpec.ts
```diff
@@ -0,0 +1,133 @@
...
+            if (treeNode[pathPart] === undefined) {
+                treeNode[pathPart] = {};
+            }
+            treeNode = treeNode[pathPart];
+        }
+        treeNode[functionName] = fn;
+    }
+    const apiType = generateApiType(publicFunctionTree);
+    const internalApiType = generateApiType(internalFunctionTree);
+    return (`
```
sshader, line 124: Could we run prettier first? I know we do this for codegen from the main convex CLI, so perhaps we can pattern match.

cli/openApiSpec.test.ts
```diff
@@ -0,0 +1,25 @@
...
+    const apiSpec = generateOpenApiSpec(JSON.parse(FUNCTIONS_JSON));
+
+    const testFileName = "openApiSpec.test.yaml";
+    fs.writeFileSync(testFileName, apiSpec, 'utf-8');
+
+    let output = execSync(`npx redocly lint ${testFileName} --format='json'`);
+    console.log(output.toString());
+    expect(output.toString()).toContain(`"errors": 0`);
+
+    fs.unlinkSync(testFileName);
```
sshader, line 18: will this cleanup still happen if the `expect` fails?

cli/openApiSpec.test.ts
```diff
@@ -0,0 +1,25 @@
...
+// If this test fails, it means the generated OpenAPI spec is no longer valid.
+test("generateValidSpec", async () => {
+    const apiSpec = generateOpenApiSpec(JSON.parse(FUNCTIONS_JSON));
+
+    const testFileName = "openApiSpec.test.yaml";
+    fs.writeFileSync(testFileName, apiSpec, 'utf-8');
+
+    let output = execSync(`npx redocly lint ${testFileName} --format='json'`);
+    console.log(output.toString());
+    expect(output.toString()).toContain(`"errors": 0`);
```
sshader, line 16: is the output not JSON? could we avoid the stringify + matching?

also ideally we'd only print the output when the test fails (I like the philosophy of passing tests printing nothing, and only failing tests printing info)

cli/openApiSpec.test.ts
```diff
@@ -0,0 +1,25 @@
...
+
+    fs.unlinkSync(testFileName);
+})
+
+test("openApiSpecMatches", () => {
```
sshader, line 21: Want to add a similar comment to the JS one?

cli/jsApiSpec.ts
```diff
@@ -0,0 +1,133 @@
+import fs from 'fs'
+import { execSync } from 'child_process';
+import { Command, Option } from "commander";
+import { ValidatorJSON } from 'convex/values'
+import { FunctionSpec } from './openApiSpec';
+import chalk from 'chalk';
+
+export const jsApiSpec = new Command("js-api-spec")
```
sshader, line 8: Should we say TS / TypeScript? I guess the TS type is the most interesting part

cli/openApiSpec.ts
```diff
@@ -0,0 +1,276 @@
...
+tags:
+    - name: query
+      description: Functions that read data
+    - name: mutation
+      description: Functions that write/update/delete data
+    - name: action
+      description: Functions that can make calls to external APIs
+paths:
+${reindent(
+        // Skip http actions because they go to a different url and we don't have argument/return types
```
sshader, line 251: Should we also filter out internal functions? Or potentially document somewhere that the internal ones need an admin key in the auth header?

cli/openApiSpec.ts
```diff
@@ -0,0 +1,276 @@
...
+            process.exit(1);
+        }
+        if (filePath) {
+            content = fs.readFileSync(filePath, 'utf-8');
+        } else {
+            const flags = options.prod ? '--prod' : '';
+            const output = execSync(`npx convex function-spec ${flags}`);
+            content = output.toString();
+        }
+        const outputPath = `convex-spec-${Date.now().valueOf()}.yaml`;
```
sshader, line 34: same comments as for the other command -- this should be configurable + the default should be documented in the help text.

cli/openApiSpec.ts
```diff
@@ -0,0 +1,276 @@
...
+                        1
+                    )}\n`
+            const propertiesStr =
+                members.length === 0
+                    ? ''
+                    : `properties:\n${reindent(members.join('\n'), 1)}`
+            return `type: object\n${requiredPropertiesStr}${propertiesStr}`
+        }
+        case 'union': {
+            const nullMember = validatorJson.value.find((v) => v.type === 'null')
```
sshader, line 118: if this is code I wrote, I wouldn't be surprised if there were some bugs around nulls in nested unions. might be worth testing a few of these manually, or we can address them as they come up

sshader:
Oh also re: package-lock.json, I think there's a particular order to run `npm install` on the different directories in this repo and if you don't do it, it results in a larger than necessary package-lock.json, which might be happening. I'll try and find some messages from Ian describing what to do

cli/openApiSpec.test.ts
jordanhunt22: not sure, i can just run it before. good call!

cli/openApiSpec.ts
jordanhunt22: okay, i can explicitly test this out more. i spent most of today converting `golinks` to use a generated clients, i confirmed that most things work correctly.

cli/openApiSpec.ts
jordanhunt22: i'll add this as a flag. we should definitely document this tho

jordanhunt22:
> Oh also re: package-lock.json, I think there's a particular order to run `npm install` on the different directories in this repo and if you don't do it, it results in a larger than necessary package-lock.json, which might be happening. I'll try and find some messages from Ian describing what to do

All your suggestions were great! I didn't realize that you could use prettier in code for formatting. Everything looks a lot better now. Please do one more pass when you get a chance. I will try to release an alpha once this is merged in.

cli/openApiSpec.ts
```diff
@@ -0,0 +1,270 @@
+import fs from 'fs'
+import { Command, Option } from "commander";
+import { ValidatorJSON } from 'convex/values'
+import chalk from 'chalk';
+import { AnalyzedFunction, FunctionSpec, getFunctionSpec } from './utils';
+import prettier from "prettier";
+
+export const openApiSpec = new Command("open-api-spec")
+    .summary("Generate an OpenAPI spec from a Convex function definition")
+    .argument("[filePath]", "The file name of the Convex function definition. If this argument is not provided, we will \nretrieve the function spec from your configured convex deployment.\n"
```
sshader, line 10: nit: get rid of the newlines?

also thoughts on `--inputFile` since it's optional? (I feel like positional args are normally required)

cli/openApiSpec.ts
```diff
@@ -0,0 +1,270 @@
...
+    .addOption(
+        new Option(
+            "--outputFile <filename>",
+            "Specify the output file name for your spec.",
+        )
+            .default(undefined)
+    )
+    .addOption(
+        new Option(
+            "--public-only",
```
sshader, line 22: Thoughts on inverting this as `--include-internal` and still defaulting to false?

cli/tsApiSpec.ts
```diff
@@ -0,0 +1,149 @@
+import fs from 'fs'
+import { Command, Option } from "commander";
+import { ValidatorJSON } from 'convex/values'
+import chalk from 'chalk';
+import { FunctionSpec, getFunctionSpec } from './utils';
+import prettier from "prettier";
+
+export const tsApiSpec = new Command("ts-api-spec")
+    .summary("Generate a JavaScript API spec from a Convex function definition")
```
sshader, line 9:
```suggestion
    .summary("Generate a TypeScript API spec similar to `convex/_generated/api.d.ts` from a Convex function definition")
```

packages/convex-helpers/README.md
```diff
@@ -773,3 +773,36 @@ import { useQuery } from "convex-helpers/react/cache";
 
 const users = useQuery(api.todos.getAll);
 ```
+
+## JavaScript API Generation
```
sshader, line 777:
```suggestion
## TypeScript API Generation
```

cli/utils.ts
```diff
@@ -0,0 +1,49 @@
...
+    if (filePath) {
+        content = fs.readFileSync(filePath, 'utf-8');
+    } else {
+        const flags = prod ? '--prod' : '';
+        let output: string;
+        try {
+            output = execSync(`npx convex function-spec ${flags}`).toString();
+        } catch (e) {
+            console.error(chalk.red("\nError retrieving function spec from your Convex deployment. " +
+                "Confirm that you \nare running this command from within a Convex project.\n"));
```
sshader, line 42: nit: get rid of some of these newlines?

sshader (approved):
Looks good to me modulo a few suggestions.

While we could bump the peer dependency requirement, that can be kinda disruptive for folks using convex helpers that don't care about this new CLI. Perhaps it's fine to just add in the error message when we shell out that you should try being on the latest convex CLI version

cli/utils.ts
jordanhunt22: It formats weirdly if I don't include the newlines, but maybe that's fine.

cli/openApiSpec.ts
jordanhunt22: that seems reasonable

cli/utils.ts
sshader: The newline after "you" seemed like it was in an odd place

PR #290, opened 2024-10-02
Triggers

ldanilek:
Implement Triggers. See the README for how to use it, and also an overview of how it works.

----

### DIFF
--- a/packages/convex-helpers/server/triggers.ts
+++ b/packages/convex-helpers/server/triggers.ts
@@ -25,7 +25,7 @@ export type Trigger<
    * This function will be called when a document in the table changes.
    * Multiple triggers on the same change will run concurrently.
    */
-  f: (ctx: Ctx, change: Change<DocumentByName<DataModel, TableName>>) => Promise<void>;
+  f: (ctx: Ctx, change: Change<DataModel, TableName>) => Promise<void>;
   /**
    * If `lock` is `true`, the trigger will be called atomically with the db change,
    * relative to other db changes and their locked triggers.
@@ -41,10 +41,14 @@ export type Trigger<
   lock?: boolean;
 };
 
-export type Change<D> = {
-  type: "create" | "update" | "delete";
-  oldDoc: D | null;
-  newDoc: D | null;
+export type Change<
+  DataModel extends GenericDataModel,
+  TableName extends TableNamesInDataModel<DataModel>,
+> = {
+  id: GenericId<TableName>;
+  type: "insert" | "update" | "delete";
+  oldDoc: DocumentByName<DataModel, TableName> | null;
+  newDoc: DocumentByName<DataModel, TableName> | null;
 };
 
 export type Triggers<Ctx, DataModel extends GenericDataModel> = {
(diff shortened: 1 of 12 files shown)


packages/convex-helpers/server/triggers.ts
```diff
@@ -0,0 +1,267 @@
...
+    const triggers = this;
+    return {
+      args: {},
+      input(ctx: Ctx) {
+        const innerDb = ctx.db;
+        return { ctx: {
+          db: new DatabaseWriterWithTriggers(ctx, innerDb, triggers),
+          innerDb,
+        }, args: {} };
+      },
```
sujayakar, line 81: you could bind `this` by doing...
```ts
input: (ctx: Ctx) => {
   ...
   new DatabaseWriterWithTriggers(ctx, innerDb, this),
},
```

packages/convex-helpers/server/triggers.ts
```diff
@@ -0,0 +1,267 @@
...
+  async lock(): Promise<(() => void)> {
+    while (this.promise !== null) {
+      await this.promise;
+    }
+    [this.promise, this.resolve] = this._newLock();
+    return () => {
+      this.promise = null;
+      this.resolve?.();
+    };
+  }
```
sujayakar, line 99: would it make sense to do a `withLock` API so it's much harder to forget to unlock?

packages/convex-helpers/server/triggers.ts
```diff
@@ -0,0 +1,267 @@
...
+ *     other triggers recursively.
+ * - Recursive triggers are queued up, so they are executed in the same order
+ *   as the database writes were. At a high level, this is a BFS traversal of
+ *   the trigger graph.
+ * - Note when there are multiple triggers, they can't be executed atomically
+ *   with the writes that caused them, from the perspective of the other
+ *   triggers. So if one trigger is making sure denormalized data is
+ *   consistent, another trigger could see the data in an inconsistent state.
+ *   To avoid such problems, triggers should be resilient to such
+ *   inconsistencies or the trigger graph should be kept simple.
```
sujayakar, line 127: do you think it'd be reasonable to just _ban_ recursive triggers for now? that way the semantics would be very straightforward. but perhaps that's too restrictive?

sujayakar (approved):
cool! some minor stuff inline but feel free to land as is.

packages/convex-helpers/server/triggers.ts
ldanilek: you actually get pretty similar guarantees when you have multiple triggers, even if they're not recursive.

like if you have these triggers:

```
triggers.register("users", async (ctx, change) => {
  const isTopStudent = await userAggregate.offsetUntil(ctx, change.newDoc.gpa) < 5;
  if (change.newDoc.isTopStudent !== isTopStudent) {
    ctx.db.patch(change.id, { isTopStudent });
  }
});
triggers.register("users", userAggregate.triggers());
```

one of the triggers has to account for the other not having run yet.

having multiple triggers seems like an important use-case, so if we have those then we might as well have recursive triggers too?

I think the semantics of recursive triggers makes sense, and I wouldn't look forward to banning them -- would it throw an error or just not execute them? -- and explaining those semantics.

PR #303, opened 2024-10-10
Change triggers API to work alongside `customCtx`

ianmacartney:
Instead of wrapping the whole custom function and needing to use its nuanced interface to extend, let's just get them using `customCtx` from day one. My prettier config breaks it into multiple lines anyways, so it wouldn't have been a one-liner either way.

I have prettier changes that I omitted for this PR. Not sure if I'm running something different or if you aren't running it

----

### DIFF
--- a/convex/triggersExample.ts
+++ b/convex/triggersExample.ts
@@ -61,13 +61,10 @@ triggers.register("counter_table", async (ctx, change) => {
   }
 });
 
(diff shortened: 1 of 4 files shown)


ldanilek (commented):
Lgtm

Please run prettier, if i didn't do it right (oop)

ldanilek (approved):

PR #326, opened 2024-10-25
Fix zodBranded conversion to Convex validator

davidoort:
See https://discord.com/channels/1019350475847499849/1298976854316814336

----

### DIFF
--- a/packages/convex-helpers/server/zod.ts
+++ b/packages/convex-helpers/server/zod.ts
@@ -36,6 +36,7 @@ import {
 } from "convex/server";
 import { Mod, NoOp, Registration } from "./customFunctions.js";
 import { pick } from "../index.js";
+import { brandedString } from "../validators.js"
 
 export type ZodValidator = Record<string, z.ZodTypeAny>;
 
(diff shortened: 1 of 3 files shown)


packages/convex-helpers/server/zod.ts
```diff
@@ -570,11 +571,8 @@ type ConvexValidatorFromZod<Z extends z.ZodTypeAny> =
                                                   ConvexValidatorFromZod<Inner>["fieldPaths"]
                                                 >
                                               : never
-                                          : Z extends z.ZodBranded<
-                                                infer Inner,
-                                                any
-                                              >
-                                            ? ConvexValidatorFromZod<Inner>
+                                          : Z extends z.ZodBranded<z.ZodString, infer Brand>
+                                            ? VString<string & { _: Brand }>
```
rjdellecese, line 575: I think you need to use the `BRAND` from `zod` (https://github.com/colinhacks/zod/blob/1f4f0dacf313a2dba45563d78171e6f016096925/src/types.ts#L4904-L4905) here:

```suggestion
                                          : Z extends z.ZodBranded<z.ZodString, infer Brand>
                                            ? VString<string & BRAND<Brand>>
```

Maybe it would be better to also support branding types other than just strings?

ianmacartney:
I cut https://github.com/davidoort/convex-helpers/pull/1 to modify this patch some - using zod brand and not modifying the `brandedString` helper

ianmacartney:
Hey @davidoort my intention with https://github.com/davidoort/convex-helpers/pull/1 was that you could merge it into your patch and try it out, then I could merge your patch in here, so you get attribution.
Are you still interested in this? 
I know you also want the zod-specific utility (that has branded input & output), but that feels independent from this PR

davidoort:
Ah right, let's leave that bonus improvement for a separate PR. I've been using `0.1.64-alpha.0` for a while now so this definitely works for me, I'll proceed with merging

davidoort:
I'll let you deal with the versioning stuff :)

PR #332, opened 2024-10-30
Write function spec to file within `api-spec` commands

jordanhunt22:
Writes the function spec to a file instead of reading from the output of a command. We have been hitting some type of buffering limit, so hopefully this resolves the issue.

----


cli/utils.ts
```diff
@@ -20,30 +22,29 @@ export type AnalyzedFunction = {
...
-        content = fs.readFileSync(filePath, 'utf-8');
-    } else {
-        const flags = prod ? '--prod' : '';
-        let output: string;
+    let filePath = filePathOpt ?? path.join(os.tmpdir(), 'function-spec.json');
+    if (!filePathOpt) {
         try {
-            output = execSync(`npx convex function-spec ${flags}`).toString();
+            const flags = prod ? "--prod" : "";
+            execSync(`npx convex function-spec ${flags} > ${filePath}`).toString();
```
thomasballinger, line 39: I think the `>` works in all the Windows shells? but it would be nice not to rely on it, instead using a flag to the `convex function-spec` command to write to a file. If `filePath` has spaces in it the code here might not work.

package.json
```diff
@@ -3,7 +3,7 @@
   "private": "true",
   "description": "Home to [packages](./packages/) to complement the official convex package.",
   "bin": {
-    "convex-helpers": "./dist/bin.cjs"
+    "convex-helpers": "./packages/convex-helpers/dist/bin.cjs"
```
thomasballinger, line 6: I don't know what these changes are doing, maybe it's confusing for this top-level package to also run the convex-helpers bin? Could you remove this? (defer to Ian)

package.json
```diff
@@ -29,8 +29,7 @@
     "@redocly/cli": "^1.20.0",
     "chalk": "^5.3.0",
     "classnames": "^2.3.2",
-    "convex": "^1.14.1",
-    "convex-helpers": "file:packages/convex-helpers/dist",
+    "convex": "^1.17.0",
```
thomasballinger, line 32: unless @ianmacartney wants this it seems unintentional

cli/utils.ts
```diff
@@ -20,30 +22,29 @@ export type AnalyzedFunction = {
     returns: ValidatorJSON | null
 };
 
-export function getFunctionSpec(prod?: boolean, filePath?: string) {
-    if (filePath && prod) {
+export function getFunctionSpec(prod?: boolean, filePathOpt?: string) {
```
thomasballinger, line 25: nit, filePathOpt and filePath aren't different enough that I know what each of them mean; does opt stand for option? Are there other names that would identify them better?

package.json
jordanhunt22: cool, i can just delete this line

cli/utils.ts
jordanhunt22: yes, opt means option. i can make this clearer

cli/utils.ts
jordanhunt22: i agree. as currently configured, there is a `--file` flag that prints out to a file. but at the end we append the timestamp, so we would have to regex the filename or something. or i can update the convex CLI code to allow you to specify a file.

not sure what your thoughts are here

thomasballinger: Ah ok the convex CLI fix doesn't have to block this fix, but not working with spaces makes this worse that it was. Especially if full paths are involved lots of folks on Windows use spaces. Instead of execSync you can use a different child_process function that allows you to direct output to a file without any shell parsing going on.

Convex CLI fix sounds useful though, it's not ideal to not be able to specify the output filename. I think I'd rather have a flag for if you need the specialized behavior of appending a timestamp? Folks using this in CI flows will have to do the same glob trick if we don't allow specifying a specific filename.

package.json
ianmacartney: yeah - the helpers import shouldn't be deleted - and you can test this yourself by running the example page from the root to verify that things work as expected.

cli/utils.ts
jordanhunt22: this seems reasonable. i am looking into the fix right now

jordanhunt22: i just increased the buffer size of `execSync`. it looks like the default is 1MB, so I increased it to 10MB. This should mitigate the issues that we are seeing because we will be able to capture the whole output.

jordanhunt22: actually, i found a way to write directly to a file instead!

thomasballinger (approved):
looks great

PR #348, opened 2024-11-20
nice error on nested triggers

ldanilek:
----

### DIFF
--- a/packages/convex-helpers/server/triggers.ts
+++ b/packages/convex-helpers/server/triggers.ts
@@ -90,6 +90,15 @@ export class Triggers<
   }
 
   wrapDB = <C extends Ctx>(ctx: C): C => {
+    if (wrapDBCalled) {
+      throw new Error(
+        `Triggers.wrapDB called multiple times in a single mutation. Not allowed due to potential deadlock.
+        Call it once in a single \`customMutation\`.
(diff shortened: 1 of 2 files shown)


packages/convex-helpers/server/triggers.ts
```diff
@@ -90,6 +90,15 @@ export class Triggers<
   }
 
   wrapDB = <C extends Ctx>(ctx: C): C => {
+    if (wrapDBCalled) {
+      throw new Error(
+        `Triggers.wrapDB called multiple times in a single mutation. Not allowed due to potential deadlock.
```
sshader, line 95: I think this could merit a specific example of what this looks like, maybe just in the triggers docs? And we could link to it from the error message?

Like
### Troubleshooting: `Triggers.wrapDB` called multiple times in a single mutation

This usually happens if a mutation is directly calling another mutation, for instance:

```
export const joinTeam = mutation(...)

export const createTeam = mutation({
	handler: (ctx) => {
		const currentUserId = // load from auth
		const team = await ctx.db.insert("teams", { creatorId: currentUserId, ... })
		// ❌ This creates a new `Triggers` since it's calling another mutation
		await joinTeam({ teamId })
	}
})
```

See <link to best practices> for the recommended pattern.

PR #349, opened 2024-11-21
Expose .handler on custom functions

ianmacartney:
Currently fails to conform to `queryGeneric` in `rowLevelSecurity.test.ts` but otherwise it'd be nice to ship it, so folks can directly call `.handler` to circumvent custom function middleware business

Can't seem to get it to satisfy MutationBuilder when I add the `& { handler }` bit...

https://github.com/get-convex/convex-helpers/blob/handler/packages/convex-helpers/server/rowLevelSecurity.test.ts#L110-L115

----

### DIFF
--- a/packages/convex-helpers/server/customFunctions.test.ts
+++ b/packages/convex-helpers/server/customFunctions.test.ts
@@ -528,6 +528,15 @@ describe("custom functions", () => {
       t.query(testApi.redefine, { a: 3 as any }),
     ).rejects.toThrow("Validator error: Expected `string`");
   });
+
+  test("inner handler is callable", async () => {
+    const t = convexTest(schema, modules);
+    expect(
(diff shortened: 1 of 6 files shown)


ianmacartney:
Instead, let's just push people towards making helpers easily:
```ts
import { ObjectType, v } from "convex/values";
import { zCustomQuery, zid, ZCustomCtx } from "convex-helpers/server/zod";

const zodQuery = zCustomQuery(query, ...);
type ZodQueryCtx = ZCustomCtx<typeof zodQuery>;

const myQueryArgs = z.object({ foo: z.number() });

async function myQueryHandler(ctx: ZodQueryCtx, args: z.infer<typeof myQueryArgs>) {
  // impl
}

export const myQuery = myZodCustomQuery({ args: myQueryArgs, handler: myQueryHandler });

PR #366, opened 2024-12-09
fix getAll to bind db

ldanilek:
fixes `getAll` so it works with stateful `db` implementations, like `WrapWriter` as used by `wrapDatabaseWriter` for RLS.

----


ldanilek:
test failures are unrelated to this PR, fixed in https://github.com/get-convex/convex-helpers/pull/367

ianmacartney (approved):

PR #367, opened 2024-12-09
fix types with new convex version

ldanilek:
whoops I changed the types in https://github.com/get-convex/convex-backend/commit/064d9d387b9d0848d4992874cd03792b9ed91646 (because `httpAction(...)` should not be callable as if it were a plain js function).

Didn't realize the type was used by HonoWithConvex.
We can fix it like this PR does, by changing the types of HonoWithConvex to use PublicHttpAction.

An alternative might be for `getRoutes` to directly return the function which has the line numbers that you want.

----

### DIFF
--- a/packages/convex-helpers/server/hono.ts
+++ b/packages/convex-helpers/server/hono.ts
@@ -104,12 +104,15 @@ export class HttpRouterWithHono<
    * @returns - an array of [path, method, endpoint] tuples.
    */
   override getRoutes = () => {
-    const convexRoutes: [string, RoutableMethod, (...args: any) => any][] = [];
+    const convexRoutes: [string, RoutableMethod, PublicHttpAction][] = [];
 
     // Likely a better way to do this, but hono will have multiple handlers with the same
     // name (i.e. for middleware), so de-duplicate so we don't show multiple routes in the dashboard.
     const seen = new Set();
     this._app.routes.forEach((route) => {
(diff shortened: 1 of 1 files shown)


packages/convex-helpers/server/hono.ts
```diff
@@ -104,20 +104,25 @@ export class HttpRouterWithHono<
...
+    const convexRoutes: [string, RoutableMethod, PublicHttpAction][] = [];
 
     // Likely a better way to do this, but hono will have multiple handlers with the same
     // name (i.e. for middleware), so de-duplicate so we don't show multiple routes in the dashboard.
     const seen = new Set();
     this._app.routes.forEach((route) => {
+      // The (internal) field _handler on PublicHttpAction is used to look up the function's line number.
+      const handler = route.handler as any;
+      handler._handler = route.handler;
+      handler.isHttp = true;
```
ianmacartney, line 116: why wouldn't it already be `isHttp`?

ianmacartney (approved):

ianmacartney:
released in 0.1.66, thanks!

packages/convex-helpers/server/hono.ts
ldanilek: It might be, but I can't tell from the type. `route.handler` has a Hono type, and as far as i know they might re-wrap the handler.

PR #368, opened 2024-12-09
await expected failures in tests

ldanilek:
Clean up warnings of the form

```
Promise returned by `expect(actual).rejects.toThrow(expected)` was not awaited. Vitest currently auto-awaits hanging assertions at the end of the test, but this will cause the test to fail in Vitest 3. Please remember to await the assertion.
```

----


ianmacartney:
ah I had this locally but hadn't pushed, thanks!

PR #385, opened 2024-12-19
fix filter types to support order

ldanilek:
the docstring of `filter` says that you can apply `filter` to queries that already have an `order`. Let's make it true.

I don't know why I need the cast `query as unknown as OrderedQuery<QueryTableInfo<Q>>`, but it looks safe and nothing is changing at runtime, only types are changing in this PR.

Added a test that previously failed.

----

### DIFF
--- a/packages/convex-helpers/server/filter.test.ts
+++ b/packages/convex-helpers/server/filter.test.ts
@@ -55,4 +55,25 @@ test("filter", async () => {
     { name: "8" },
     { name: "9" },
   ]);
+
+  // Check that ordering works, before or after the filter.
+  const withOrder = await t.run((ctx) =>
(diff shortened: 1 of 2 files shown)


ianmacartney (approved):
So the ordered query type allows both passing in a non-ordered or ordered query? I'm guessing this means you could accidentally put it in both places and types wouldn't complain?

ldanilek:
> So the ordered query type allows both passing in a non-ordered or ordered query? I'm guessing this means you could accidentally put it in both places and types wouldn't complain?

`filter` doesn't change the type: `filter(query)` has the same type as `query`. So if you put order inside, `query.order()` is an OrderedQuery so `filter(query.order())` is an ordered query and `filter(query.order()).order()` will complain

PR #440, opened 2025-02-07
fix paginator endCursor when isDone

ldanilek:
there's a bug where `endCursor` is not used for the last page of pagination. What should happen is the last page grows to include everything up to endCursor, like every other page. But what was happening instead is it was being cut off at num-items.

----


ianmacartney (commented):
👍

PR #446, opened 2025-02-08
composable streams of query results

ldanilek:
as described in the README, add some helpers for merging and filtering streams of query results.

This extracts the implementations of `paginator` and `streamQuery` into new helpers, that allow better composition and more patterns.

First there's `reflect` which allows you to construct queries with the normal syntax `reflect(ctx.db, schema).query(table).withIndex(index, indexRange).order("desc")` and then get their internal details, e.g. which index it's looking at. -- the naming is based on https://golangbot.com/reflection/ but it's mostly an internal library.

Then once you have a reflectable query, it can be used as a "stream", which is an async iterable of documents ordered by an index. For this reason, `stream(ctx.db, schema)` is another way of writing `reflect(ctx.db, schema)`.

Once you have a stream, you can merge them with `mergeStreams` or filter them with `filterStream`, generating more streams.

See the README for more details and motivating examples.

----


packages/convex-helpers/server/stream.ts
```diff
@@ -0,0 +1,1001 @@
...
+    narrow: (indexBounds: IndexBounds) => {
+      return concatStreams(...streams.map(stream => stream.narrow(indexBounds)));
+    },
+  };
+}
+
+/**
+ * Apply a filter to a stream.
+ * 
+ * Watch out for sparse filters, as they may read unbounded amounts of data.
```
ianmacartney, line 758: Could we pass in a `rowReadLimit` or something so folks can bound - like "give me 10 non-deleted users' messages, but cap the search at 1k users. Would that end up with surprising results later? where pagination might think it's done? Maybe throw a catch-able error with the results so far? Or do you have other ideas for how to gracefully fail there? Maybe it's "usually" totally fine, but there was a burst of user deletions, and now a query is stuck in a failing state?

ianmacartney (commented):
I think I get it, but the language around queries, query streams, index streams, and merging is a bit murky even for me after skimming the code. Can you give some bullets of how each term relates to one another in the readme? 

I'm starting to wonder if all the query helpers could get bundled into a `@convex-dev/query-utils` library that you own. It might be a good organization, and more "official-feeling" than helpers. I've already pulled migrations and rate limiting into components, and maybe I should pull validators into `validator-utils` or something, then have helpers be for developing APIs before crystalizing them into separate packages... one downside of course is discoverability.

ianmacartney:
tbh I hit some review fatigue when I got to the huge stream.ts and compare.ts logic. I can look more closely at whatever feels most important to you, or go deep into it once I have self-hosting things under control

packages/convex-helpers/README.md
```diff
@@ -853,6 +853,101 @@ export const list = query({
...
+import schema from "./schema";
+
+export const latestMessages = query({
+  args: { author: v.id("users") },
+  handler: async (ctx, { author }) => {
+    const messagesForUnreadStatus = [false, true].map(unread => stream(ctx.db, schema).query("messages")
+      .withIndex("by_author", q => q.eq("author", author).eq("unread", unread))
+      .order("desc"));
+    const allMessagesByCreationTime = mergeStreams(...messagesForUnreadStatus);
+    return await queryStream(allMessagesByCreationTime).take(10);
```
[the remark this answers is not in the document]
ldanilek: mergeStreams combines the given streams into a single stream, sorted by the index key. so you don't need to sort (i should make this more explicit in the readme)

packages/convex-helpers/server/compare.ts
```diff
@@ -0,0 +1,86 @@
...
+  }
+  if (v instanceof ArrayBuffer) {
+    return [6, Array.from(new Uint8Array(v)).map(makeComparable)];
+  }
+  if (Array.isArray(v)) {
+    return [7, v.map(makeComparable)];
+  }
+  // Otherwise, it's an POJO.
+  const keys = Object.keys(v).sort();
+  const pojo: Value[] = keys.map((k) => [k, v[k]!]);
```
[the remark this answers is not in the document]
ldanilek: i don't think they're the same thing

packages/convex-helpers/server/compare.ts
```diff
@@ -0,0 +1,86 @@
+import { Value } from "convex/values";
```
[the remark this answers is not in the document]
ldanilek: i thought to do that i had to make an export in package.json
(i don't mind making this an entrypoint, but it shouldn't be necessary for now)

packages/convex-helpers/server/stream.ts
```diff
@@ -0,0 +1,1001 @@
...
+    }
+    this.lowerBoundIndexKey = this.lowerBoundIndexKey ?? [];
+    this.lowerBoundIndexKey.push(value);
+    this.upperBoundIndexKey = this.upperBoundIndexKey ?? [];
+    this.upperBoundIndexKey.push(value);
+    return this;
+  }
+  lt(field: string, value: Value) {
+    if (!this.canUpperBound(field)) {
+      throw new Error(`Cannot use lt on field '${field}'`);
```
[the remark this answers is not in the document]
ldanilek: this is like if you have an index on `["a", "b"]` and you do `q=>q.lt("b", 1)`, or `q=>q.lt("a", 1).lt("a", 1)` or `q=>q.eq("a", 1).lt("a", 1)`. there are tests :)

packages/convex-helpers/server/stream.ts
```diff
@@ -0,0 +1,1117 @@
...
+}
+
+/**
+ * A "stream" is an async iterable of query results, ordered by an index on a table.
+ */
+export interface IndexStream<
+  DataModel extends GenericDataModel,
+  T extends TableNamesInDataModel<DataModel>,
+> {
+  iterWithKeys(): AsyncIterable<[DocumentByName<DataModel, T> | null, IndexKey]>;
```
ldanilek, line 208: TODO: IndexKey can contain undefined, make sure we're not serializing it as a convex value

packages/convex-helpers/README.md
```diff
@@ -853,6 +853,151 @@ export const list = query({
...
+get a stream of "messages created by user1 or user2 that are unread". Then you
+can convert the stream into a query and paginate it.
+
+Concrete functions you can use:
+
+- `stream` constructs a stream using the same syntax as `DatabaseReader`.
+  - e.g. `stream(ctx.db, schema).query("messages").withIndex("by_author", (q) => q.eq("author", "user1"))`
+- `mergeStreams` combines multiple streams into a new stream, ordered by the same index.
+- `filterStream` filters out documents from a stream based on a TypeScript predicate.
+- `queryStream` converts a stream into a query, so you can call `.first()`, `.collect()`, `.paginate()`, etc.
```
ldanilek, line 880: necessary?

packages/convex-helpers/README.md
```diff
@@ -853,6 +853,151 @@ export const list = query({
 });
 ```
 
+## Composable streams of query results
+
+These are helper functions for constructing and composing streams of query results.
+
```
ldanilek, line 859: add a motivating example here

packages/convex-helpers/README.md
```diff
@@ -853,6 +853,151 @@ export const list = query({
...
+export const latestMessages = query({
+  args: { author: v.id("users") },
+  handler: async (ctx, { author }) => {
+    // These are two streams of messages, ordered by _creationTime descending.
+    // The first has read messages, the second has unread messages.
+    const messagesForUnreadStatus = [false, true].map((unread) =>
+      stream(ctx.db, schema)
+        .query("messages")
+        .withIndex("by_author", (q) =>
+          q.eq("author", author).eq("unread", unread),
```
ldanilek, line 989: does this order by author,unread still? or by _creationTime?

packages/convex-helpers/README.md
```diff
@@ -853,6 +853,151 @@ export const list = query({
...
+        )
+        .order("desc"),
+    );
+    // Merge the two streams into a single stream of all messages authored by
+    // `args.author`, ordered by _creationTime descending.
+    const allMessagesByCreationTime = mergeStreams(...messagesForUnreadStatus);
```
ldanilek, line 995: separate cursor for each mergeStream?

ldanilek:
@ianmacartney I think i addressed all of the feedback from the code review and our in-person chat. I'd like to land it now and we can iterate afterwards.
(in general i like smaller PRs; this one kind of got out of hand, and i can make adjustments on top of this incrementally)

PR #447, opened 2025-02-08
Generate proper types for a Record

nyoung697:
Generate the correct types for a record, instead of just returning any.

The type `RecordKeyValidatorJSON` isn't exported from the convex package, but if it were to be exported, the `generateRecordKeyType` function can be updated.


packages/convex-helpers/cli/tsApiSpec.ts
```diff
@@ -99,6 +102,24 @@ function generateArgsType(argsJson: ValidatorJSON): string {
...
+ */
+function generateRecordKeyType(keys: any): string {
+    switch (keys.type) {
+        case 'string':
+            return 'string'
+        case 'id':
+            return `Id<"${keys.tableName}">`
+        case 'union':
+            return keys.value.map(generateRecordKeyType).join(' | ')
+        default:
```
jordanhunt22, line 118: this default should be `any`

packages/convex-helpers/cli/tsApiSpec.ts
```diff
@@ -76,8 +76,11 @@ function generateArgsType(argsJson: ValidatorJSON): string {
             return `Id<"${argsJson.tableName}">`
         case 'array':
             return `Array<${generateArgsType(argsJson.value)}>`
-        case 'record':
-            return 'any'
+        case 'record': {
+            const keyType = generateRecordKeyType(argsJson.keys)
+            const valueType = generateArgsType(argsJson.values.fieldType)
```
jordanhunt22, line 81: this `value` type is also restricted because it can't be optional

jordanhunt22 (commented):
Thanks @nyoung697 for this contribution!

I'm going to update the `convex` package to export the `RecordKeyValidatorJSON` and `RecordValueValidatorJSON` objects. That way, we can do this correctly. Can you please ping me in 1-2 weeks if I don't circle back?

packages/convex-helpers/cli/tsApiSpec.ts
nyoung697: Thanks Jordan! What do you mean by this?

nyoung697: I modified it to check if the field is optional and it will add undefined. Is this the correct way of handling this?

```
case 'record': {
      const keyType = generateRecordKeyType(argsJson.keys);
      const valueType = generateArgsType(argsJson.values.fieldType);
      const isOptional = argsJson.values.optional ? " | undefined" : "";
      return `Record<${keyType}, ${valueType}${isOptional}>`;
    }
```

packages/convex-helpers/cli/tsApiSpec.ts
```diff
@@ -106,6 +110,24 @@ function generateArgsType(argsJson: ValidatorJSON): string {
   }
 }
 
+/**
+ * Generates a TypeScript-compatible key type for a record.
+ *
+ * The keys should actually be RecordKeyValidatorJSON, but this isn't exported from validators.ts in convex.
```
jordanhunt22, line 116: Can you please update this to use the types I exported [here](https://github.com/get-convex/convex-js/commit/32e0386e51da407477e250f1db2a5fa1aff3037b)? Please also do the same with the values.

packages/convex-helpers/cli/tsApiSpec.ts
```diff
@@ -83,8 +83,12 @@ function generateArgsType(argsJson: ValidatorJSON): string {
       return `Id<"${argsJson.tableName}">`;
     case "array":
       return `Array<${generateArgsType(argsJson.value)}>`;
-    case "record":
-      return "any";
+    case "record": {
+      const keyType = generateRecordKeyType(argsJson.keys);
+      const valueType = generateArgsType(argsJson.values.fieldType);
+      const isOptional = argsJson.values.optional ? " | undefined" : "";
```
jordanhunt22, line 89: values should never be optional

ianmacartney (changes requested):
see jordan's feedback

nyoung697:
I added the `RecordKeyValidatorJSON` type. I don't know what you want me to do with the value type.

jordanhunt22 (approved):
These changes look good. Thanks for contributing! Can you confirm that we can modify, copy, and redistribute this contribution by adding a thumbs up to this message?

Once that's done, we can go ahead and get this merged in.

ianmacartney:
It's live in `convex-helpers@0.1.76` - thanks for your patience!

PR #484, opened 2025-03-06
Ian/cors no clobber

### LINKED ISSUE #376: corsRouter can absorb non-cors routes at the same path
If you do this:
```ts
const http = httpRouter();
const cors = corsRouter(http);

http.route({
  path: "/foo",
  method: "POST",
...
});
cors.route({
  path: "/foo",
  method: "GET",
...
});
export default http;
```
Then the `OPTIONS` endpoint will say that both `POST` and `GET` are allowed, even though only the latter is. It won't return CORS headers from the `POST` endpoint, but still may mislead browsers.

It's also funky in that this does work:
```ts
const http = httpRouter();
const cors = corsRouter(http);

cors.route({
  path: "/foo",
  method: "GET",
...
});
http.route({
  path: "/foo",
  method: "POST",
...
});
export default http;
```

We should keep track of the methods for each path, not just slurp up every defined method at a path.

ianmacartney:
Fixes #376 

----

### DIFF
--- a/packages/convex-helpers/server/cors.test.ts
+++ b/packages/convex-helpers/server/cors.test.ts
@@ -1,4 +1,4 @@
-import { describe, test, expect, vi } from "vitest";
+import { describe, test, expect, vi, assert } from "vitest";
 import { corsRouter } from "./cors";
 import { defineSchema, httpActionGeneric, HttpRouter } from "convex/server";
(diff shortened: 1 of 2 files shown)


mikecann (approved):
Looks good to me. Tested locally and seems to work as expected

PR #490, opened 2025-03-08
add missing case for paginated queries

PierreCapo:
It seems useStablePaginatedQuery is missing a loading case to not rerender the UI.
This PR add the missing case and ensures the UI is not rerendering while loading the first page. 

It's particularly visible when a paginated query is used with a `withSearchIndex` and a user is typing a text that does a paginated query. Without this check, the UI keeps rerendering every time the user types on a character on the keyboard.

----


ianmacartney (approved):

ianmacartney:
Thank you for catching this!

PR #495, opened 2025-03-12
improve .unique() error message

djbalin:
_Similar to [this PR I opened](https://github.com/get-convex/convex-backend/pull/59) in the convex-backend repo_

Makes the error message thrown by .unique() more useful. Now, the _id's of the first two documents returned by the unique() query are printed in the error message. This is very helpful to the user in identifying and resolving incorrect data.

The wording in the error message has been changed such that it is identical to the error thrown by `.unique()` from `convex-backend` (slightly confusing that the same error has different error messages based on where it's thrown from).

Feel free to change my wording in the error message - but in my opinion it should be clear that the document id's printed are not all the documents returned by the unique() query but just the first two.

**Note:** I did not build and test this change locally, and intellisense did not help me in verifying that `_id` does indeed exist on those objects, so please verify before merging!

----

### DIFF
--- a/packages/convex-helpers/server/filter.ts
+++ b/packages/convex-helpers/server/filter.ts
@@ -27,7 +27,7 @@ import {
 
 async function asyncFilter<T>(
   arr: T[],
-  predicate: (d: T) => Promise<boolean> | boolean,
+  predicate: (d: T) => Promise<boolean> | boolean
 ): Promise<T[]> {
   const results = await Promise.all(arr.map(predicate));
   return arr.filter((_v, index) => results[index]);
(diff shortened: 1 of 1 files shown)


packages/convex-helpers/server/filter.ts
```diff
@@ -84,7 +84,9 @@ class QueryWithFilter<T extends GenericTableInfo>
       if (uniqueResult === null) {
         uniqueResult = result;
       } else {
-        throw new Error("not unique");
+        throw new Error(
+          `unique() query returned more than one result. First pair of results: ${uniqueResult._id} and ${result._id}`
+        );
```
nipunn1313, line 89: nice! Can you do the same technique from https://github.com/get-convex/convex-backend/pull/59#pullrequestreview-2679927198 to make it terser?

ianmacartney (approved):

PR #497, opened 2025-03-14
zod returns, object, and more

### LINKED ISSUE #70: zodToConvexFields doesn't return the correct type when using zod transforms
When creating table fields using zod, using a transform on a zod schema results in:
1. The return value of the transform is not used as the schema return type - instead the input type is always used
2. Intellisense/autocomplete breaks when trying to get schema fields from the `getEntDefinitions` helper

Example implementation/use case:

```js
import { defineEnt, defineEntSchema, getEntDefinitions } from "convex-ents";
import { zodToConvexFields } from "convex-helpers/server/zod";
import { z } from "zod";

const message = {
  sentTimestamp: z.string().transform((val) => new Date(val).getTime()),
  // ^ sentTimestamp: v.string()
  // expected: v.number()
};

const schema = defineEntSchema({
  messages: defineEnt(zodToConvexFields(message)),
});

const definitions = getEntDefinitions(schema);

type SentTimestamp = typeof definitions.messages.document.sentTimestamp;
// ^ "Value" / no type
// expected: number
// also no intellisense/autocomplete after typing "document."

// Zod is able so succesfully return the correct type for a string input transformed into a number
const result = z.object(message).parse({ sentTimestamp: "2024-01-01T00:00:00Z" });
type SentTimestamp = typeof result.sentTimestamp;
// ^ number (expected)
```

ianmacartney:
- Allows you to specify a `returns` validator on the zod custom query that will become a Convex validator as well, allowing inferring the api return types via tools like `ts-api-spec`. This replaces the `output` handler, which will still works but will not set the convex validator, out of an abundance of caution for backwards compatibility.
- Allows you to skip convex validation altogether if you're using zod validation with `skipConvexValidation: true` on the custom query. This is per-query site right now, but could be a global config if folks want it.
- Allows passing a `z.object` as args, or an inline `{}` as returns. It does not currently support unions for args.
- Provide a utility `zodOutputToConvex` and corresponding type `ConvexValidatorFromZodOutput<Z>` that produce a validator of the zod validator's *output* type, which differs for default values, pipelines, and effects. Note: effects don't have a well-defined output validator, so it's validated as `any` (which doesn't allow undefined).

Fixes #70 

----

### DIFF
--- a/convex/_generated/api.d.ts
+++ b/convex/_generated/api.d.ts
@@ -14,7 +14,6 @@ import type {
   FunctionReference,
 } from "convex/server";
 import type * as addIt from "../addIt.js";
-import type * as corsHttpRouterExample from "../corsHttpRouterExample.js";
 import type * as counter from "../counter.js";
 import type * as http from "../http.js";
 import type * as migrationsExample from "../migrationsExample.js";
@@ -36,7 +35,6 @@ import type * as triggersExample from "../triggersExample.js";
  */
 declare const fullApi: ApiFromModules<{
   addIt: typeof addIt;
-  corsHttpRouterExample: typeof corsHttpRouterExample;
   counter: typeof counter;
   http: typeof http;
   migrationsExample: typeof migrationsExample;
--- a/packages/convex-helpers/server/zod.test.ts
+++ b/packages/convex-helpers/server/zod.test.ts
@@ -126,20 +126,23 @@ export const failsReturnsValidator = zQuery({
 });
 
 export const zodOutputCompliance = zQuery({
-  args: {},
-  handler: async () => {
+  // Note no args validator
+  handler: (ctx, args: { maybe?: string | undefined }) => {
     return {
       default: undefined,
       effect: "effect",
       pipeline: 3,
       extraArg: "extraArg",
+      maybe: args.maybe,
     };
   },
-  returns: v.object({
+  // Note inline record of zod validators works.
+  returns: {
     default: z.string().default("default"),
     effect: z.string().transform((s) => null),
     pipeline: z.number().pipe(z.coerce.string()),
-  }),
+    maybe: z.string().optional(),
+  },
 });
 
 /**
@@ -532,6 +535,18 @@ test("zod output compliance", async () => {
     effect: null,
     pipeline: "3",
   });
+  const responseWithMaybe = await t.query(testApi.zodOutputCompliance, {
+    maybe: "maybe",
+  });
+  expect(responseWithMaybe).toMatchObject({
+    maybe: "maybe",
+  });
+  // number should fail
+  await expect(() =>
+    t.query(testApi.zodOutputCompliance, {
+      maybe: 1,
+    }),
+  ).rejects.toThrow();
 });
 
 describe("zod functions", () => {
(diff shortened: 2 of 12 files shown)


convex/_generated/api.d.ts
```diff
@@ -14,7 +14,6 @@ import type {
   FunctionReference,
 } from "convex/server";
 import type * as addIt from "../addIt.js";
-import type * as corsHttpRouterExample from "../corsHttpRouterExample.js";
```
sshader, line 17: hmm test that codegen stays up to date for this repo could be useful

packages/convex-helpers/server/zod.test.ts
```diff
@@ -486,6 +529,45 @@ test("zod date round trip", async () => {
...
+  ).rejects.toThrow();
+});
+
+test("zod output compliance", async () => {
+  const t = convexTest(schema, modules);
+  const response = await t.query(testApi.zodOutputCompliance, {});
+  expect(response).toMatchObject({
+    default: "default",
+    effect: null,
+    pipeline: "3",
```
sshader, line 545: nit: perhaps some more descriptive names on the fields could make it easier to read this test without having to scroll back and forth to the implementation. like `args.numberArg` and `response.stringifiedNumberArg`

packages/convex-helpers/server/zod.ts
```diff
@@ -393,24 +440,50 @@ export type CustomBuilder<
...
             ...args: OneOrZeroArgs extends [infer A]
               ? [Expand<A & ModMadeArgs>]
               : [ModMadeArgs]
           ) => ReturnValue;
-        }
+          /**
+           * If true, the function will not be validated by Convex,
+           * in case you're seeing performance issues with validating twice.
+           */
+          skipConvexValidation?: boolean;
```
sshader, line 466: Does this one actually make sense with `output`? Or is it simpler to include it on both in case someone wants to add it everywhere

packages/convex-helpers/server/zod.ts
```diff
@@ -784,6 +861,327 @@ export function zodToConvex<Z extends z.ZodTypeAny>(
...
+      return v.any() as ConvexValidatorFromZodOutput<Z>;
+    case "ZodArray":
+      const inner = zodOutputToConvex(zod._def.type);
+      if (inner.isOptional === "optional") {
+        throw new Error("Arrays of optional values are not supported");
+      }
+      return v.array(inner) as ConvexValidatorFromZodOutput<Z>;
+    case "ZodObject":
+      return v.object(
+        zodToConvexFields(zod._def.shape()),
```
sshader, line 1095: does this not need to go down the output specific pathway? Looks like this calls `zodToConvex` internally (idk if you can do things like nest a default within an object)

sshader (approved):
want to double check the `zodToConvexFields` usage, but otherwise LGTM

packages/convex-helpers/server/zod.ts
ianmacartney: With output it would still disable the input validation happening with Convex in addition to zod

packages/convex-helpers/server/zod.ts
ianmacartney: great catch! I missed that - `zodToConvexFields` doesn't have an equivalent right now. I just added tests and updated this. Maybe I should just make `zodOutputToConvexFields` for completeness..

ianmacartney (commented):
thanks for catching that 😅

PR #574, opened 2025-05-10
useSessionPaginatedQuery

thomasballinger:
Is that what we want to name it? I'll publish an alpha now


ianmacartney (approved):
lovely

ianmacartney:
shipped in `convex-helpers@0.1.86`

PR #576, opened 2025-05-12
Allow ConvexHttpClient in withArgs

thomasballinger:
As a future direction how about withArgs can return more full wrappers of various clients? If that works as a plan then it's fine to add more clients here and a type parameter for the client type in the future, but for now it's just the same interface!

### DIFF
--- a/packages/convex-helpers/browser.ts
+++ b/packages/convex-helpers/browser.ts
@@ -1,5 +1,5 @@
 import type { BetterOmit, EmptyObject } from "./index.js";
-import type { ConvexClient } from "convex/browser";
+import type { ConvexClient, ConvexHttpClient } from "convex/browser";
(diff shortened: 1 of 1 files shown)


nipunn1313 (approved):

ianmacartney:
If you mean having them implement more of the interface, yes sgtm - I was wondering if this is the right spot for a proxy object approach.

If what you mean by full client is that this does the initialization too, then I'm not sure.
For stateful clients, I suspect there will be a lot of hybrid calls with and without extra params. And I suspect this one could be nested too to have e.g. one that supplies an apiKey argument, and another elsewhere that adds a userId argument?

thomasballinger:
Proxy or manually passing through the supported methods/properties sounds good, makes sense that these should be lightweight wrappers so you can create many.

Issue #558, opened 2025-04-30
Update zod validation to support Zod 4


thebruceyang:
I ran the zod 4 update docs against server/zod. Hope this helps (at least logs) some required updates:
 
### Key Compatibility Issues:
1. **Internal Structure Access**: 
   The library heavily relies on accessing `._def` directly:
   ```typescript
   const typeName: ZodFirstPartyTypeKind | "ConvexId" = zod._def.typeName;
   ```
   But the Zod 4 migration guide explicitly states:
   ```
   moves ._def
   The ._def property is now moved to ._zod.def.
   ```
   This is a critical issue that would prevent the library from working correctly with Zod 4.

2. **Function Returns API**:
   The library does have some backward compatibility:
   ```typescript
   let returns = fn.returns ?? fn.output;
   ```
   It checks for both `returns` and `output`, which is good, but the core function API in Zod 4 has changed significantly.

3. **Schema Access Pattern**:
   Many of the helper functions like `zodToConvex` and `zodOutputToConvex` directly access internal properties of Zod schemas that have changed in Zod 4.

### What This Means:

1. This version of convex-helpers is **not compatible** with Zod 4 without modifications.

2. You would need:
   - An updated version of convex-helpers specifically built for Zod 4
   - To modify the library code to use `._zod.def` instead of `._def`
   - To update any other internal structure access patterns

3. The `returns` field mentioned in the previous document might work because the library explicitly checks for both `returns` and `output`, but the underlying validation mechanism would still fail due to the internal structure changes.

ianmacartney:
I'm going to wait until it's out of beta to make any changes, and I wonder
if the zod4 version should get published as its own library - I've wondered
if @convex-dev/zod` would be a good move - to start separating helpers
into first-class packages

ianmacartney:
Feel free to start a branch that makes the changes based on the migration
guide though! That'd be a big help

ksanfi:
Probably relevant https://github.com/colinhacks/zod/issues/4371

ianmacartney:
Totally! I really like the approach and I think it would make sense to support this in a similar way with both a convex/helpers/zod3 and convex/helpers/zod4 endpoint that where (for now) the bare /zod endpoint exports everything from /zod3

The reason I'm hesitant to support both in the same functions is everyone would end up with twice as much code as they need. With tree shaking it can only import what it needs. 

Anyone brave enough to start the zod4 implementation? I learned a lot about Zod making the last one. 
[...]

PR #593, opened 2025-05-22
Add standard schema support

ianmacartney:
not quite #558 but at least you can do something like 

```ts
import { toStandardSchema } from "convex-helpers/standardSchema";

const standardValidator = toStandardSchema(v.object({
  name: v.string(),
  age: v.number(),
}));

standardValidator["~standard"].validate({
  name: "John",
  age: 30,
});
```

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ find the npm package in [./packages/convex-helpers](./packages/convex-helpers).
 | [OpenAPI Spec Generator](./packages/convex-helpers/README.md#open-api-spec-generation)
 | [Triggers](./packages/convex-helpers/README.md#triggers)
 | [CORS for HttpRouter](./packages/convex-helpers/README.md#cors-support-for-httprouter)
+| [Standard Schema support](./packages/convex-helpers/README.md#standard-schema)
 
 | In this directory for copy-pasting:
 | -----------------------------------
--- a/package-lock.json
+++ b/package-lock.json
@@ -8209,7 +8209,7 @@
     },
     "packages/convex-helpers/dist": {
       "name": "convex-helpers",
-      "version": "0.1.88-alpha.0",
+      "version": "0.1.88",
       "license": "Apache-2.0",
       "bin": {
         "convex-helpers": "bin.cjs"
--- a/package-lock.json
+++ b/package-lock.json
@@ -8209,7 +8209,7 @@
     },
     "packages/convex-helpers/dist": {
       "name": "convex-helpers",
-      "version": "0.1.88",
+      "version": "0.1.89-alpha.0",
       "license": "Apache-2.0",
       "bin": {
         "convex-helpers": "bin.cjs"
--- a/package-lock.json
+++ b/package-lock.json
@@ -8209,7 +8209,7 @@
     },
     "packages/convex-helpers/dist": {
       "name": "convex-helpers",
-      "version": "0.1.87",
+      "version": "0.1.88-alpha.0",
       "license": "Apache-2.0",
       "bin": {
         "convex-helpers": "bin.cjs"
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,6 +19,7 @@
         "@arethetypeswrong/cli": "0.18.1",
         "@edge-runtime/vm": "5.0.0",
         "@redocly/cli": "1.34.3",
+        "@standard-schema/spec": "^1.0.0",
         "@testing-library/react": "16.3.0",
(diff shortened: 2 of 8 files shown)


natedunn:
A few notes on me playing around with this:

### 1. I was unaware that `.validate` isn't typesafe 

```ts
const vSchema = toStandardSchema(
	v.object({
		name: v.string(),
	})
)["~standard"].validate({
	name: 1234, // Does not typecheck in IDE
});

const zSchema = z
	.object({
		name: z.string(),
	})
["~standard"].validate({
	name: 1234, // Also doesn't typecheck in IDE
});
```

### 2. They don't return the same types

I am unsure of the implications here but I noticed they return differently

```
const vSchema: StandardSchemaV1.Result<Value> | Promise<StandardSchemaV1.Result<Value>>
```

vs. 

```
const zSchema: StandardSchemaV1.Result<{
    name: string;
}> | Promise<StandardSchemaV1.Result<{
    name: string;
}>>
```

### 3. Unable to use with standard schema accepted APIs

When using Tanstack Form to test, `toStandardSchema` version passes with acceptable types, however it cannot actually run validation. Example of a Tanstack Start/router route: 

```ts
import { createFileRoute } from "@tanstack/react-router";
import { v } from "convex/values";
import { toStandardSchema } from "convex-helpers/standardSchema";
import z from "zod";
import { useForm } from "@tanstack/react-form";

export const Route = createFileRoute("/_default/test")({
	component: RouteComponent,
});

function RouteComponent() {
	const vSchema = toStandardSchema(
		v.object({
			name: v.string(),
		})
	)["~standard"].validate;

	const zSchema = z.object({
		name: z.string(),
	});

	const form = useForm({
		defaultValues: {
			name: "",
		},
		validators: {
			// onSubmit: zSchema, // This one works
			onSubmit: vSchema, // This does not work
		},
		onSubmit: async ({ value }) => console.log(value),
	});

	return (
		<div className="container">
			<form
				onSubmit={(e) => {
					e.preventDefault();
					e.stopPropagation();
					form.handleSubmit();
				}}
			>
				<form.Field name="name">
					{(field) => (
						<div className="flex gap-3 items-end">
							<div className="flex flex-col gap-2 flex-1">
								<label className="label">
									<span className="label-text">Name</span>
								</label>
								<input
									className="input border border-black"
									name={field.name}
									value={field.state.value}
									onChange={(e) => {
										console.log(e.target.value);
										field.handleChange(e.target.value);
									}}
								/>
							</div>
						</div>
					)}
				</form.Field>
				<form.Subscribe
					selector={(state) => [state.canSubmit, state.isSubmitting]}
				>
					{([isSubmitting]) => (
						<button type="submit">
							Submit
						</button>
					)}
				</form.Subscribe>
				<pre>
					{JSON.stringify(
						form.state.errors.map((err) => err),
						null,
						2
					)}
				</pre>
			</form>
		</div>
	);
}
```

### 4. Unsure of the value

I am unsure of the value this brings, at this point. As discussed in this [Discord thread](https://discord.com/channels/1019350475847499849/1374833143902765107), I think the real value is either developing a helper that takes any Standard Schema validation library and converting it into Convex's `values` syntax OR allowing args and `defineTable`/ `defineSchema` to accept Standard Schema libs out of the box. 

Since those options seem to have significant hurdles, I also brought up a method of potentially adding a refinement function that could take in an existing schema and add additional validation refinements from there. See [Drizzle's refinement approach](https://orm.drizzle.team/docs/zod#refinements) for this inspiration.

ianmacartney:
1. Yeah zod does this too - I think so you can pass an `unknown` value in and get a typed one out.
2. Good catch! Fixed in `convex-helpers@0.1.89-alpha.0`
3. I'm hoping that is just a matter of changing:
```diff
 const vSchema = toStandardSchema(
	v.object({ name: v.string() })
-)["~standard"].validate;
+);

const zSchema = z.object({
	name: z.string(),
});
```

natedunn:
With `convex-helpers@0.1.89-alpha.0` we are cooking with # 3 in the list

```ts
const vSchema = toStandardSchema(
	v.object({
		name: v.string(),
	})
);

const form = useForm({
	defaultValues: {
		name: "",
	},
	validators: {
		onSubmit: vSchema, 
	},
	onSubmit: async ({ value }) => console.log(value),
});
```
Confirmed with non string types as well

CleanShot 2025-05-23 at 12  31 31

ianmacartney:
great! Sounds like it's all set, so it's now out in `convex-helpers@0.1.89`

Issue #622, opened 2025-06-02
Parse throws with valid data with unions of literals


ThomasKientz:
parse method throws unexpectedly on valid values:

```ts
const validator = v.union(v.literal('foo'), v.literal('bar'))

parse(validator, 'foo') // returns 'foo'
parse(validator, 'bar') // throws "Validator error: Expected `literal`, got `string`"
```

`parse(validator, 'bar')` should not throw as "bar" is a valid value from union.

ianmacartney:
ah, good call - there's a bug with `validate` if you pass `{ throw: true }`, which `parse` does. I'll have a fix out shortly. 
thanks for the report. cc @davidoort

ianmacartney:
in `convex-helpers@0.1.90`

PR #623, opened 2025-06-02
fix parse to handle union of literal types

ThomasKientz:
fixes https://github.com/get-convex/convex-helpers/issues/622


ianmacartney:
Your fix works - I'm going to make it throw a more informative error - right now it will say something like 
```
Validator error: Expected `union`, got `string`
```

ianmacartney:
in `convex-helpers@0.1.90`

PR #634, opened 2025-06-05
fix stream pagination for descending cursors

### LINKED ISSUE #630: Stream/paginator fails descending pagination from a cursor
When doing descending cursor-based pagination with one item at a time, this data:
```
f1 | f2
 0 | 0
 0 | 1
 1 | 0
 1 | 1
```
ordered by [f1, f2]
will return `1, 1` then `0, 1` then throw an error that `1, 0` is out of order, when called with a cursor at each step.

Interestingly, when iterated in ascending order, it'll be
```
0, 0 <- first request
0, 1 <- second request
```
then the next request will return both `1, 0` and `1, 1` at once, then the next pages (of 1) will be empty.

This may be an artifact of having a page size of 1 in this test, but is breaking reliably for the Agent component. 

I'll get a minimal repro together to help debug / write a failing test.

ianmacartney:
Fixes #630

We were producing ranges to query that were like
a lt 2
a eq 2, b lt 3
a eq 2, b eq 3, c lt 4
...

but for descending iteration, we need these ranges to be reversed, which is important for ConcatStreams

----

### DIFF
--- a/convex/counter.ts
+++ b/convex/counter.ts
@@ -1,3 +1,4 @@
+import { paginationOptsValidator } from "convex/server";
 import { action, query } from "./_generated/server";
 // Using mutation from triggersExample so any changes will run triggers
 import { mutation } from "./triggersExample";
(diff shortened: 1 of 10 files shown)


ianmacartney:
in `convex-helpers@0.1.91-alpha.0` to test

PR #638, opened 2025-06-09
Fix types on customFunctions

benjavicente:
Note: Currently, this PR only adds missing test that fail type checking but correctly passes tests.

---

Custom functions are incorrectly typed because the given query/mutation/action to extend is assumed that has the same type definition as the base convex function builders. This is inconvenient when a custom function want's to extend another, like in the new added tests.

### DIFF
--- a/packages/convex-helpers/server/customFunctions.test.ts
+++ b/packages/convex-helpers/server/customFunctions.test.ts
@@ -367,7 +367,7 @@ export const outerExtender = customQuery(inner, {
 export const outerExtends = outerExtender({
   args: { a: v.string() },
   handler: async (ctx, args) => {
-    return { outer: ctx.outer, ...args };
+    return { outer: ctx.outer, inner: ctx.inner, ...args };
   },
 });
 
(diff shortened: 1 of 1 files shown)


benjavicente:
I couldn't understand the types, I will try later but would happily accept if someone continues this PR  😵‍💫

---

This are my notes:

```ts
// 1. When `inner` (the first argument of a custom query) is `CustomBuilder` (function override?)

// 2. Take the inner context, input parameters and added parameters
inner: CustomBuilder<"query", InnerArgsValidator, InnerCtx, InnerMadeArgs, GenericQueryCtx<DataModel>, Visibility>

// 3. Pass the resulting context from inner?
mod: Mod<
  InnerCtx,
  // NEW?: what args inner took, added to the function args.
  // Each element here can't be present on outer args unless added on InnerMadeArgs.
  InnerArgsValidator,
  // NEW?: what args inner added, added as "default" args.
  // Those args will be propagated down to outer unless consumed in InnerArgsValidator.
  InnerMadeArgs;
  ModArgsValidator extends PropertyValidators,
  ModCtx extends Record<string, any>,
  ModMadeArgs extends Record<string, any>,
>
```

ianmacartney:
Yes I haven't figured out whether TypeScript's generics can actually
support doing the nesting and having the types flow through without
changing the API.
This is why the current recommendation is to only have one level and share
code within them - e.g. a function that adds something to ctx, or makes a
bunch of shared things like loggers

PR #664, opened 2025-06-25
Custom function allows extra params

ianmacartney:
Now you can specify extra parameters at the function definition site and use them in the customFunction builder

```ts
const myQueryBuilder = customQuery(query, {
  args: {  },
  input: async (ctx, args, { role }: { role: "admin" | "user" }) => {
    const user = await getUser(ctx);
    if (role === "admin" && user.role !== "admin") {
      throw new Error("You are not an admin");
    }
    if (role === "user" && !user) {
      throw new Error("You must be logged in to access this query");
    }
    return { ctx: { user }, args: {} };
  }
});

const myAdminQuery = myQueryBuilder({
  role: "admin",
  args: { },
  handler: async (ctx, args) => {
    // ...
  },
});
```

----

### DIFF
--- a/convex/testingFunctions.ts
+++ b/convex/testingFunctions.ts
@@ -21,11 +21,14 @@ export const testingQuery = customQuery(query, {
 
 export const testingMutation = customMutation(mutation, {
   args: {},
-  input: async (_ctx, _args) => {
+  input: async (_ctx, _args, { devOnly }: { devOnly: boolean }) => {
     if (process.env.IS_TEST === undefined) {
       throw new Error(
         "Calling a test only function in an unexpected environment",
       );
+    }
+    if (devOnly && process.env.IS_PROD) {
+      throw new Error("This function is only available in development");
     }
     return { ctx: {}, args: {} };
   },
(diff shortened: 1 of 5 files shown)


hipdev:
🚀

ianmacartney:
@hipdev want to try it out? `npm i https://pkg.pr.new/get-convex/convex-helpers@664` hopefully works - just set that up.
Would love any feedback before this becomes canon

hipdev:
Trying it tonight!

ianmacartney:
Did you get a chance to try it yet?

hipdev:
@ianmacartney
It’s working perfectly! It also saves me a lot of code and makes the mutations safer, since in some of them I was sending data directly from the client just to quickly prototype.

This is what I’m doing in a new `builders.ts` file, and it’s working perfectly!

image

ianmacartney:
> @ianmacartney It’s working perfectly! It also saves me a lot of code and makes the mutations safer, since in some of them I was sending data directly from the client just to quickly prototype.
> 
> This is what I’m doing in a new `builders.ts` file, and it’s working perfectly!
> 
> image

@hipdev from the screenshot it doesn't look like you're using the extra arguments feature? I'd expect there to be a third argument to `input: async (ctx, args, extra: { ...yourtypes })`

Just curious if you've tried that part yet?

PR #678, opened 2025-07-06
feat(crud): Add support for union types and complex objects

front-depiction:
### **What Changed**
Enhanced the `crud()` function to properly support union types and complex nested objects with comprehensive test coverage.

#### **Fixed Validator Composition**
create a new validator with fields that are recursively made optional. This enables the removal of the validator.kind must be object condition.

#### **New Recursive Validator Utilities**
- **`makeOptional`** - Makes all fields optional for update patches
- **`makeSystemFieldsOptional`** - Adds optional system fields for create operations

Both functions recursively handle:
- Objects → Field-level transformations
- Unions → Applied to each member
- Records → Applied to value types
- Others → Optional wrapping

###  **Test Coverage**
Added **9 new test cases** covering:

**Union Tables**
- User type: `{type: "user", name: string, email: string}`
- Admin type: `{type: "admin", name: string, permissions: string[]}`
- Guest type: `{type: "guest", sessionId: string}`

**Complex Objects**
- Nested objects with arrays and optional fields
- Partial updates with proper validation
- Optional field handling

**Edge Cases**
- Pagination with cursors
- Delete operations returning documents
- Non-existent document handling

###  **Benefits**
Enables CRUD operations with any table schema including polymorphic data, complex nested structures, and mixed validation scenarios.

### DIFF
--- a/packages/convex-helpers/server/crud.ts
+++ b/packages/convex-helpers/server/crud.ts
@@ -20,11 +20,10 @@ import {
 import type {
   GenericId,
   Infer,
-  OptionalProperty,
   Validator,
 } from "convex/values";
 import { v } from "convex/values";
-import { partial, systemFields } from "../validators.js";
+import { doc, partial } from "../validators.js";
 /**
  * Create CRUD operations for a table.
  * You can expose these operations in your API. For example, in convex/users.ts:
@@ -73,62 +72,42 @@ export function crud<
 ) {
   type DataModel = DataModelFromSchemaDefinition<SchemaDefinition<Schema, any>>;
 
-  const validator = schema.tables[table]?.validator;
+  const validator = schema.tables[table]?.validator
   if (!validator) {
     throw new Error(
       `Table ${table} not found in schema. Did you define it in defineSchema?`,
     );
   }
 
-  const makeOptional = <
-    Type,
-    IsOptional extends OptionalProperty,
-    FieldPaths extends string,
-  >(
-    validator: Validator<Type, IsOptional, FieldPaths>,
-  ): Validator<Type, "optional", FieldPaths> => {
-    if (validator.kind === "object") {
-      return v.object(partial(validator.fields)) as any;
-    } else if (validator.kind === "union") {
-      return v.union(
-        ...validator.members.map((value) => makeOptional(value) as any),
-      ) as any;
-    } else {
-      return v.optional(validator) as Validator<Type, "optional", never>;
-    }
-  };
+  const systemFields = v.object({
+    _id: v.id(table),
+    _creationTime: v.number(),
+  });
 
-  const optionalValidator = makeOptional(validator);
+  const partialSystemFields = partial(systemFields).fields;
 
-  const makeSystemFieldsOptional = <
-    Type,
-    IsOptional extends OptionalProperty,
-    FieldPaths extends string,
-  >(
-    validator: Validator<Type, IsOptional, FieldPaths>,
-  ): Validator<Type, IsOptional, FieldPaths> => {
+
+  const makeSystemFieldsOptional = (
+    validator: Validator<any, any, any>,
+  ): Validator<any, any, any> => {
     if (validator.kind === "object") {
       return v.object({
         ...validator.fields,
-        ...partial({
-          _id: v.optional(v.id(table)),
-          _creationTime: v.optional(v.number()),
-        }),
+        ...partialSystemFields,
       }) as any;
     } else if (validator.kind === "union") {
       return v.union(
-        ...validator.members.map((value) => makeSystemFieldsOptional(value)),
+        ...validator.members.map((value) => makeSystemFieldsOptional(value) as any),
       ) as any;
     } else {
       return validator;
     }
   };
 
-  const validatorWithSystemFields = makeSystemFieldsOptional(validator);
 
   return {
     create: mutation({
-      args: validatorWithSystemFields,
+      args: makeSystemFieldsOptional(validator),
       handler: async (ctx, args) => {
         if ("_id" in args) delete args._id;
         if ("_creationTime" in args) delete args._creationTime;
@@ -172,7 +151,7 @@ export function crud<
         id: v.id(table),
         // this could be partial(table.withSystemFields) but keeping
         // the api less coupled to Table
-        patch: optionalValidator,
+        patch: partial(v.union(doc(schema, table)))
       },
       handler: async (ctx, args) => {
         await ctx.db.patch(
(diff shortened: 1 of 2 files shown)


packages/convex-helpers/server/crud.ts
```diff
@@ -67,28 +67,44 @@ export function crud<
...
-    throw new Error(
-      `CRUD only supports simple tables ${table} is a ${validator.type}`,
-    );
-  }
+
+
+
+  const makeOptional: (validator: Validator<any, any, any>) => Validator<any, any, any> =
+    (validator: Validator<any, any, any>) =>
+      "fields" in validator
```
ianmacartney:
```suggestion
      validator.kind === "object"
```

packages/convex-helpers/server/crud.ts
```diff
@@ -67,28 +67,44 @@ export function crud<
...
+        ? v.object(partial(validator.fields))
+        : "members" in validator
+          ? v.union(...validator.members.map((value) => makeOptional(value)))
+          : validator.kind === "record" ? v.record(validator.key, makeOptional(validator.value))
+            : v.optional(validator);
```
ianmacartney:
```suggestion
          : v.optional(validator);
```
We don't support top-level records, and top-level fields will get replaced, not recursively updated by `db.patch`

packages/convex-helpers/server/crud.ts
```diff
@@ -67,28 +67,44 @@ export function crud<
...
+      "fields" in validator
+        ? v.object({
+          ...validator.fields,
+          ...(validator.fields._id ? { _id: v.optional(validator.fields._id) } : {}),
+          ...(validator.fields._creationTime ? { _creationTime: v.optional(validator.fields._creationTime) } : {}),
+        })
+        : "members" in validator
+          ? v.union(...validator.members.map((value) => makeSystemFieldsOptional(value)))
+          : validator.kind === "record" ? v.record(validator.key, makeSystemFieldsOptional(validator.value))
+            : validator;
```
ianmacartney: This will not do anything currently, since `validator` does not have system fields set on it

packages/convex-helpers/server/crud.ts
```diff
@@ -67,28 +67,44 @@ export function crud<
...
+  const optionalValidator = makeOptional(validator);
+  const createValidator = makeSystemFieldsOptional(validator);
 
   return {
     create: mutation({
-      args: {
-        ...validator.fields,
-        ...partial(systemFields),
-      },
+      args: createValidator,
```
ianmacartney: one thing to validate is whether the inferred types come out correct - if the `makeSystemFieldsOptional` results in a `Validator<any, any, any>` then you lose type safety when calling `create`

ianmacartney (changes requested):
Thanks! 
One reason I was doing the simpler approach was to allow types to flow through. There is also a `systemFields(table)` helper I believe, which could simplify too. 
Let's scope this to adding union support and keep the more straightforward system field treatment

packages/convex-helpers/server/crud.ts
front-depiction: The typing situation I am rather unfamiliar with. I do not know what the exact intended behavior of the validators especially for the union type.

packages/convex-helpers/server/crud.ts
```diff
@@ -67,28 +72,37 @@ export function crud<
...
+  const makeOptional = <
+    Type,
+    IsOptional extends OptionalProperty,
+    FieldPaths extends string,
+  >(
+    validator: Validator<Type, IsOptional, FieldPaths>,
+  ): Validator<Type, "optional", never> => {
+    if (validator.kind === "object") {
+      return v.object(partial(validator.fields)) as any;
+    } else if ("members" in validator) {
```
ianmacartney:
```suggestion
    } else if (validator.kind === "union") {
```

packages/convex-helpers/server/crud.ts
```diff
@@ -67,28 +72,37 @@ export function crud<
...
+
+  const optionalValidator = makeOptional(validator);
 
   return {
     create: mutation({
-      args: {
-        ...validator.fields,
-        ...partial(systemFields),
-      },
+      args: validator,
```
ianmacartney: For historical reasons, we don't fail if the user passes in system fields here or to patch, even though they'll get ignored. So we still need to include them in the validator as optional, so easiest is likely to revert to the `...partial(systemFields)` handling IMO

packages/convex-helpers/server/crud.ts
```diff
@@ -67,28 +72,37 @@ export function crud<
...
-    );
-  }
+
+  const makeOptional = <
+    Type,
+    IsOptional extends OptionalProperty,
+    FieldPaths extends string,
+  >(
+    validator: Validator<Type, IsOptional, FieldPaths>,
+  ): Validator<Type, "optional", never> => {
```
ianmacartney:
```suggestion
  ): Validator<Type, "optional", FieldPaths> => {
```

packages/convex-helpers/server/crud.ts
front-depiction: can't do ...validator.fields on a union. That's the reason behind the previous recursive method to make the system field optional even on more complex objects

packages/convex-helpers/server/crud.ts
```diff
@@ -67,28 +72,63 @@ export function crud<
...
+
+  const makeOptional = <
+    Type,
+    IsOptional extends OptionalProperty,
+    FieldPaths extends string,
+  >(
+    validator: Validator<Type, IsOptional, FieldPaths>,
+  ): Validator<Type, "optional", FieldPaths> => {
+    if (validator.kind === "object") {
+      return v.object(partial(validator.fields)) as any;
```
ianmacartney: the return type here is "optional" but it's returning a required `v.object`. I think the return type should be whatever it was before / probably needs to be required as the body of a table, and the return type should probably be `Partial<Type>`.

This was harder than it should be, so I just added support for `partial(validator)` without splitting out `.fields` or `.members` when the validator is an object or union in the latest version of helpers. 
You can see the implementation here:
https://github.com/get-convex/convex-helpers/blob/main/packages/convex-helpers/validators.ts#L122-L144

I suspect doing `partial(doc(schema, table))` will do the trick!

front-depiction: partial(doc(schema, table)) does not work, as the input types don't match up. slapping it in a union does the trick but feels rather hacky.

ianmacartney (approved):
Great, thanks for making these changes! Putting it out in the next alpha now

PR #692, opened 2025-07-18
Fix incorrect type inference: ensure ZodString instead of generic ZodType<string> on convexToZod function types

### LINKED ISSUE #671: ZOD convexToZod better types
I propose better type for convexToZod Function that make zod types less general and extendible like  z.string().min(1) or  z.number().max(100).min(0) and something else...

Personally I use it for client side form validation having the database as one source of true

```
type ZodFromValidatorBase<V extends GenericValidator> =
    V extends VId<GenericId<infer TableName>> ? Zid<TableName> :
    V extends VString<any, any> ? z.ZodString :
    V extends VFloat64<any, any> ? z.ZodNumber :
    V extends VInt64<any, any> ? z.ZodBigInt :
    V extends VBoolean<any, any> ? z.ZodBoolean :
    V extends VNull<any, any> ? z.ZodNull :
    V extends VLiteral<any, any> ? z.ZodLiteral<V["value"]> :
    V extends VObject<any, any, any, any>
    ? z.ZodObject<{ [K in keyof V["fields"]]: ZodValidatorFromConvex<V["fields"][K]> }> :
    V extends VRecord<any, any, any, any, any>
    ? z.ZodRecord<z.ZodString, ZodValidatorFromConvex<V["value"]>> :
    V extends VArray<any, any>
    ? z.ZodArray<ZodValidatorFromConvex<V["element"]>> :
    V extends VUnion<any, any, any, any>
    ? z.ZodUnion<[ZodValidatorFromConvex<V["members"][number]>]>
    : z.ZodTypeAny; // fallback for unknown validators



// Main type with optional handling
export type ZodValidatorFromConvex<V extends GenericValidator> =
    V extends Validator<any, "optional", any>
    ? z.ZodOptional<ZodFromValidatorBase<V>>
    : ZodFromValidatorBase<V>;



export function convexToZod<V extends GenericValidator>(
    convexValidator: V,
): ZodValidatorFromConvex<V> {
    const isOptional = convexValidator.isOptional === "optional";

    let zodValidator//: z.ZodTypeAny;

    switch (convexValidator.kind) {
        case "id":
            zodValidator = zid((convexValidator as VId<any>).tableName);
            break;
        case "string":
            zodValidator = z.string();
            break;
        case "float64":
            zodValidator = z.number();
            break;
        case "int64":
            zodValidator = z.bigint();
            break;
        case "boolean":
            zodValidator = z.boolean();
            break;
        case "null":
            zodValidator = z.null();
            break;
        case "any":
            zodValidator = z.any();
            break;
        case "array": {
            const arrayValidator = convexValidator as VArray<any, any>;
            zodValidator = z.array(convexToZod(arrayValidator.element));
            break;
        }
        case "object": {
            const objectValidator = convexValidator as VObject<any, any>;
            zodValidator = z.object(convexToZodFields(objectValidator.fields));
            break;
        }
        case "union": {
            const unionValidator = convexValidator as VUnion<any, any, any, any>;
            const memberValidators = unionValidator.members.map(
                (member: GenericValidator) => convexToZod(member),
            );
            zodValidator = z.union([
                memberValidators[0],
                memberValidators[1],
                ...memberValidators.slice(2),
            ]);
            break;
        }
        case "literal": {
            const literalValidator = convexValidator as VLiteral<any>;
            zodValidator = z.literal(literalValidator.value);
            break;
        }
        case "record": {
            const recordValidator = convexValidator as VRecord<
                any,
                any,
                any,
                any,
                any
            >;
            zodValidator = z.record(
                convexToZod(recordValidator.key),
                convexToZod(recordValidator.value),
            );
            break;
        }
        default:
            throw new Error(`Unknown convex validator type: ${convexValidator.kind}`);
    }

    return isOptional ? z.optional(zodValidator) as ZodValidatorFromConvex<V> : zodValidator as ZodValidatorFromConvex<V>;
}

export function convexToZodFields<C extends PropertyValidators>(
    convexValidators: C,
) {
    return Object.fromEntries(
        Object.entries(convexValidators).map(([k, v]) => [k, convexToZod(v)]),
    ) as { [k in keyof C]: ZodValidatorFromConvex<C[k]>//z.ZodType<Infer<C[k]>>
        };
}

```

Firephoenix25:
## What This PR Does

This PR fixes an issue where the inferred type of a Zod schema is too generic (`ZodType<string, ZodTypeDef, string>`), which prevents chaining string-specific methods such as `.min()`, `.email()`, etc.

## Problem

When conditionally constructing a schema, the inferred type ends up being the generic `ZodType<string, ZodTypeDef, string>` instead of the more specific `ZodString`. As a result, methods like `.min()` are not recognized by TypeScript:

Closes #671 

I have read the CLA Document and I hereby sign the CLA.

### DIFF
--- a/packages/convex-helpers/server/zod.ts
+++ b/packages/convex-helpers/server/zod.ts
@@ -1323,8 +1323,10 @@ export type ConvexToZod<V extends GenericValidator> = z.ZodType<Infer<V>>;
 type ZodFromValidatorBase<V extends GenericValidator> =
   V extends VId<GenericId<infer TableName extends string>>
     ? Zid<TableName>
-    : V extends VString<any, any>
-      ? z.ZodString
+    : V extends VString<infer T, any>
+      ? T extends string & { _: infer Brand extends string }
+        ? z.ZodBranded<z.ZodString, Brand>
+        : z.ZodString
       : V extends VFloat64<any, any>
         ? z.ZodNumber
         : V extends VInt64<any, any>
@@ -1333,24 +1335,22 @@ type ZodFromValidatorBase<V extends GenericValidator> =
             ? z.ZodBoolean
             : V extends VNull<any, any>
               ? z.ZodNull
-              : V extends VLiteral<any, any>
-                ? z.ZodLiteral<V["value"]>
-                : V extends VObject<any, any, any, any>
-                  ? z.ZodObject<{
-                      [K in keyof V["fields"]]: ZodValidatorFromConvex<
-                        V["fields"][K]
-                      >;
-                    }>
-                  : V extends VRecord<any, infer Key, any, any, any>
+              : V extends VLiteral<infer T, any>
+                ? z.ZodLiteral<T>
+                : V extends VObject<any, infer Fields, any, any>
+                  ? z.ZodObject<
+                      {
+                        [K in keyof Fields]: ZodValidatorFromConvex<Fields[K]>;
+                      },
+                      "strip"
+                    >
+                  : V extends VRecord<any, infer Key, infer Value, any, any>
                     ? Key extends VId<GenericId<infer TableName>>
                       ? z.ZodRecord<
                           Zid<TableName>,
-                          ZodValidatorFromConvex<V["value"]>
+                          ZodValidatorFromConvex<Value>
                         >
-                      : z.ZodRecord<
-                          z.ZodString,
-                          ZodValidatorFromConvex<V["value"]>
-                        >
+                      : z.ZodRecord<z.ZodString, ZodValidatorFromConvex<Value>>
                     : V extends VArray<any, any>
                       ? z.ZodArray<ZodValidatorFromConvex<V["element"]>>
                       : V extends VUnion<any, any, any, any>
--- a/packages/convex-helpers/server/zod.test.ts
+++ b/packages/convex-helpers/server/zod.test.ts
@@ -833,6 +833,9 @@ test("convexToZod complex types", () => {
 
   const recordValidator = convexToZod(v.record(v.string(), v.number()));
   expect(recordValidator.constructor.name).toBe("ZodRecord");
+
+  const optionalValidator = convexToZod(v.optional(v.string()));
+  expect(optionalValidator.constructor.name).toBe("ZodOptional");
 });
 
 test("convexToZodFields", () => {


packages/convex-helpers/server/zod.ts
```diff
@@ -1314,14 +1314,65 @@ export function zBrand<
...
+      : V extends VFloat64<any, any>
+        ? z.ZodNumber
+        : V extends VInt64<any, any>
+          ? z.ZodBigInt
+          : V extends VBoolean<any, any>
+            ? z.ZodBoolean
+            : V extends VNull<any, any>
+              ? z.ZodNull
+              : V extends VLiteral<any, any>
+                ? z.ZodLiteral<V["value"]>
```
ianmacartney: Could this be `VLiteral<infer T, any>` instead of using `V["value"]`?

packages/convex-helpers/server/zod.ts
```diff
@@ -1314,14 +1314,65 @@ export function zBrand<
...
+                : V extends VObject<any, any, any, any>
+                  ? z.ZodObject<{
+                      [K in keyof V["fields"]]: ZodValidatorFromConvex<
```
ianmacartney: Same here for extracting the type instead of using `V["fields"]` - though maybe your version avoids some inference challenges I'm unaware of? wdyt?

packages/convex-helpers/server/zod.ts
```diff
@@ -1314,14 +1314,65 @@ export function zBrand<
...
+ *
+ * ES: z.ZodString instead of z.ZodType<string, z.ZodTypeDef, string>
+ * so you can use methods of z.ZodString like .min() or .email()
+ */
+
+type ZodFromValidatorBase<V extends GenericValidator> =
+  V extends VId<GenericId<infer TableName extends string>>
+    ? Zid<TableName>
+    : V extends VString<any, any>
+      ? z.ZodString
```
ianmacartney: Do we lose branded types here? `e.g. the `brandedString` helper results in a `VString<string & {_brand: "foo" },` type - can we pass that through to a more specific `ZodString` type?

packages/convex-helpers/server/zod.ts
```diff
@@ -1393,7 +1444,9 @@ export function convexToZod<V extends GenericValidator>(
       throw new Error(`Unknown convex validator type: ${convexValidator.kind}`);
   }
 
-  return isOptional ? z.optional(zodValidator) : zodValidator;
+  return isOptional
+    ? (z.optional(zodValidator) as ZodValidatorFromConvex<V>)
```
ianmacartney, line 1448: are optional types going through? I fear above that we're losing that.

Firephoenix25: I think yes because I'm extracing it from Validator<any, "optional", any> that is a generic validator

ianmacartney (approved):
Have you installed the package version from pkg-pr-new and it's been working for you?

ianmacartney:
I cut an alpha with the change - it'll go into the next main release, thanks for your work here!

PR #693, opened 2025-07-21
fix(triggers): Make DatabaseWriterWithTriggers methods enumerable for…

front-depiction:
… spreading

The methods of `DatabaseWriterWithTriggers` were not enumerable, preventing them from being included when the `db` context was spread (e.g., `{...ctx.db}`). This caused issues when attempting to wrap or log the context in downstream middleware.

By explicitly binding and assigning these methods as instance properties in the constructor, they become enumerable. This ensures compatibility with operations like object spreading, matching the expected behavior of the original Convex `db` object.

This change specifically addresses `get`, `query`, `normalizeId`, `insert`, `patch`, `replace`, and `delete` methods.

----

### DIFF
--- a/packages/convex-helpers/server/triggers.ts
+++ b/packages/convex-helpers/server/triggers.ts
@@ -171,6 +171,14 @@ export class DatabaseWriterWithTriggers<
     this.#triggers = triggers;
     this.#isWithinTrigger = isWithinTrigger;
     this.system = innerDb.system;
+      // Make methods enumerable by assigning them as instance properties
+    this.get = this.get.bind(this);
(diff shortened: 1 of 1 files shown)


ianmacartney:
Huh I've never seen this pattern before for classes...
I cut a branch making a first-class object - would you be willing to try it out? #695 there should be a release in there shortly that you can install

front-depiction:
Yes all works. I had initially just opted for a binding of the methods to keep the change minimal. Object is still the better move imo to better mimic the expected behavior from the original context.

ianmacartney:
Great! It's in the latest alpha

PR #706, opened 2025-07-29
fix union types was wrong

Firephoenix25:
```
const result = v.union(
    v.object({
        kind: v.literal("StringDocument"),
        value: v.string(),
    }),
    v.object({
        kind: v.literal("NumberDocument"),
        value: v.number(),
    }),
);
```

this code should result with comma separated object instead of | (vertical bar) like that👇🏻

```
z.ZodUnion<[z.ZodObject<{
    kind: z.ZodLiteral<"StringDocument">;
    value: z.ZodString;
}, "strip", z.ZodTypeAny, {
    value: string;
    kind: "StringDocument";
}, {
    value: string;
    kind: "StringDocument";
}>, z.ZodObject<{
    kind: z.ZodLiteral<"NumberDocument">;
    value: z.ZodNumber;
}, "strip", z.ZodTypeAny, {
    value: number;
    kind: "NumberDocument";
}, {
    value: number;
    kind: "NumberDocument";
}>]>
```

instead of 

```
z.ZodUnion<[z.ZodObject<{
    kind: z.ZodLiteral<"StringDocument">;
    value: z.ZodString;
}, "strip", z.ZodTypeAny, {
    value: string;
    kind: "StringDocument";
}, {
    value: string;
    kind: "StringDocument";
}> | z.ZodObject<{
    kind: z.ZodLiteral<"NumberDocument">;
    value: z.ZodNumber;
}, "strip", z.ZodTypeAny, {
    value: number;
    kind: "NumberDocument";
}, {
    value: number;
    kind: "NumberDocument";
}>]>
```

----

### DIFF
--- a/packages/convex-helpers/server/zod.ts
+++ b/packages/convex-helpers/server/zod.ts
@@ -1353,9 +1353,26 @@ type ZodFromValidatorBase<V extends GenericValidator> =
                       : z.ZodRecord<z.ZodString, ZodValidatorFromConvex<Value>>
                     : V extends VArray<any, any>
                       ? z.ZodArray<ZodValidatorFromConvex<V["element"]>>
-                      : V extends VUnion<any, any, any, any>
+                      : V extends VUnion<
+                            any,
+                            [
+                              infer A extends GenericValidator,
+                              infer B extends GenericValidator,
+                              ...infer Rest extends GenericValidator[],
+                            ],
+                            any,
+                            any
+                          >
                         ? z.ZodUnion<
-                            [ZodValidatorFromConvex<V["members"][number]>]
+                            [
+                              ZodValidatorFromConvex<A>,
+                              ZodValidatorFromConvex<B>,
+                              ...{
+                                [K in keyof Rest]: ZodValidatorFromConvex<
+                                  Rest[K]
+                                >;
+                              },
+                            ]
                           >
                         : z.ZodTypeAny; // fallback for unknown validators
 


packages/convex-helpers/server/zod.ts
```diff
@@ -1353,9 +1353,26 @@ type ZodFromValidatorBase<V extends GenericValidator> =
                       : z.ZodRecord<z.ZodString, ZodValidatorFromConvex<Value>>
                     : V extends VArray<any, any>
                       ? z.ZodArray<ZodValidatorFromConvex<V["element"]>>
-                      : V extends VUnion<any, any, any, any>
+                      : V extends VUnion<
+                            any,
+                            [
+                              infer A extends GenericValidator,
+                              infer B extends GenericValidator,
```
ianmacartney, line 1360: We used to enforce that there were at least two args to VUnion, but we no longer. So I suspect we'd want to do the `...infer` without these
```suggestion
```

Firephoenix25: but z.ZodUnion require at least 2 elements

ianmacartney: Ah, gotcha. Thanks. This looks great!

Firephoenix25: maybe I have to throw an error in the function saying that zod support minimum 2 elements on the union or something like that

ianmacartney:
If zod supports it at runtime, I think we're all set. If zod fails at
runtime, then that error will suffice.
I guess I didn't test `convexToZod(v.union(v.null()))` - not sure if that
has a type error for not having multiple items

PR #709, opened 2025-07-31
Use compareValues from convex-helpers

Nicolapps:
This fixes an issue with people using older convex NPM versions that don't include compareValues


Nicolapps:
It looks like the implementation in the Convex NPM package has some fixes the old convex-helpers implementation misses

ianmacartney:
It sounds like we should bump up the peer dependency for convex helpers to
a minimum version of when compareValues came out?

Nicolapps:
@ianmacartney Yes, I think this would make sense. This issue has affected one of our customers

PR #712, opened 2025-08-01
Export QueryStream

APZelos:
I'm creating Effect wrappers around Convex and ConvexHelpers, and while working on the stream helpers, I struggled a lot with TypeScript because I don't have access to the type of the internal `QueryStream` class. If it's ok to export it that would help me avoid having to do castings to `any` 😅.
(If not, that's fine)

----


ianmacartney (approved):
makes sense! thanks

ianmacartney:
you can test that it fixes it with the pkg-pr-new version above. If that's all good I'll put it out in the next release

z4p5a9:
> you can test that it fixes it with the pkg-pr-new version above. If that's all good I'll put it out in the next release

Just tested it, works great! Thanks a lot 🙌

PR #720, opened 2025-08-05
Add option for default deny behavior to RLS Helpers

ebg1223:
Tl;dr- adds a config option to RLS helpers to specify default deny behavior, instead of current default allow behavior.

Prior to this PR, using row level security helpers would always allow a request if a function was not explicitly provided for a method on a table. 

This PR adds a config type RLSConfig, with a single option, defaultPolicy, which accepts either "allow" or "deny". 
The wrapDatabaseReader and wrapDatabaseWriter optionally accept an object of this type. 

If no config object is provided, the default behavior remains the same, and the request will be allowed if no method handler for the table is provided.
If {defaultPolicy: "deny"} is provided, access will be denied, unless a method handler for the table is provided, and returns true.

Tests also added to validate behavior.

Thank you for reading!

----

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice. - I Agree

### DIFF
--- a/packages/convex-helpers/server/rowLevelSecurity.test.ts
+++ b/packages/convex-helpers/server/rowLevelSecurity.test.ts
@@ -1,10 +1,11 @@
 import { convexTest } from "convex-test";
 import { v } from "convex/values";
 import { describe, expect, test } from "vitest";
-import { wrapDatabaseWriter } from "./rowLevelSecurity.js";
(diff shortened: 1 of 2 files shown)


ianmacartney:
out in convex-helpers@0.1.104

PR #731, opened 2025-08-18
fix: make all partial args result in optional arg

Yash-Singh1:
This PR makes it so that functions that have all of their args set to optional don't have to pass an empty object to the session query to fix type errors.

----


ianmacartney:
The version above should be good to validate ^.
I'll merge it and release once I have a chance to try it. but you can use the version above in the meantime

Yash-Singh1:
Just tested it with the snapshot you provided. I can confirm that it's working
Screenshot 2025-08-18 at 5 01 40 PM

ianmacartney:
nice work!

PR #818, opened 2025-10-15
Zod 4 Support

### LINKED ISSUE #558: Update zod validation to support Zod 4

### LINKED ISSUE #793: Zid can't be in a z.array
The Zid type of the Zod helpers doesn't seem to implement whatever is necessary to be able to belong to a z.array.

The specific error I'm getting is:

```
lib/validation/competitions.ts:23:21 - error TS2345: Argument of type 'Zid<"genres">' is not assignable to parameter of type 'SomeType'.
  Property '_zod' is missing in type 'Zid<"genres">' but required in type 'SomeType'.

23   genreIds: z.array(zid("genres")),
                       ~~~~~~~~~~~~~

  node_modules/zod/v4/core/schemas.d.cts:86:5
    86     _zod: _$ZodTypeInternals;
           ~~~~
    '_zod' is declared here.
```

ksinghal:
### Summary
- This is an initial effort towards implementing zod4 support.
- Add Zod v4 monolithic integration for Convex in `packages/convex-helpers/server/zod4.ts`.
- Strongly typed implementation (no loose any for public APIs); lints clean.

### Key Changes
- New API surface in `server/zod4.ts`: `zid`, `zBrand`, `zCustomQuery/Mutation/Action`, `withSystemFields`, `zodToConvex`, `zodToConvexFields`, `zodOutputToConvex`, `convexToZod`, `toConvexJS`, `fromConvexJS`, `convexCodec`.
- Package export: add `./server/zod4` subpath (types + default) in `packages/convex-helpers/package.json`.

### Documentation
- Expanded Zod Validation section for v4 in `packages/convex-helpers/README.md` (see README for usage and examples).

### Tests
- Added smoke test `packages/convex-helpers/server/zod4.test.ts` (parity suite to follow).

### Peer Dependency
- Update `zod` peer range to `^3.22.4 || ^4.1.12` (v4.1+ required).

### Compatibility
- Existing v3 helper (`server/zod`) remains unchanged; new v4 functionality is opt-in via `./server/zod4`.

### Follow-ups
- Port full v3 parity tests to `zod4.test.ts`.
- Add an end-to-end example module using `server/zod4`.

Fixes #558 
Fixes #793

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -226,7 +226,6 @@
       "integrity": "sha512-2BCOP7TN8M+gVDj7/ht3hsaO/B/n5oDbiAyyvnRlNOs+u1o+JWNYTQrmpuNp1/Wq2gcFrI01JAW+paEKDMx/CA==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.27.1",
         "@babel/generator": "^7.28.3",
@@ -816,7 +815,6 @@
         }
       ],
       "license": "MIT",
-      "peer": true,
       "engines": {
         "node": ">=18"
       },
@@ -840,7 +838,6 @@
         }
       ],
       "license": "MIT",
-      "peer": true,
       "engines": {
         "node": ">=18"
       }
@@ -861,7 +858,6 @@
       "integrity": "sha512-NKBGBSIKUG584qrS1tyxVpX/AKJKQw5HgjYEnPLC0QsTw79JrGn+qUr8CXFb955Iy7GUdiiUv1rJ6JBGvaKb6w==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@edge-runtime/primitives": "6.0.0"
       },
@@ -2020,7 +2016,6 @@
       "integrity": "sha512-dKYCMuPO1bmrpuogcjQ8z7ICCH3FP6WmxpwC03yjzGfZhj9fTJg6+bS1+UAplekbN2C+M61UNllGOOoAfGCrdQ==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@octokit/auth-token": "^4.0.0",
         "@octokit/graphql": "^7.1.0",
@@ -2234,7 +2229,6 @@
       "integrity": "sha512-3giAOQvZiH5F9bMlMiv8+GSPMeqg0dbaeo58/0SlA9sxSqZhnUtxzX9/2FzyhS9sWQf5S0GJE0AKBrFqjpeYcg==",
       "dev": true,
       "license": "Apache-2.0",
-      "peer": true,
       "engines": {
         "node": ">=8.0.0"
       }
@@ -3031,8 +3025,7 @@
       "resolved": "https://registry.npmjs.org/@standard-schema/spec/-/spec-1.0.0.tgz",
       "integrity": "sha512-m2bOd0f2RT9k8QJx1JN85cZYyH1RqFBdlwtkSlf4tBDYLCiiZnv1fIIwacK6cqwXavOydf0NPToMQgpKq+dVlA==",
       "devOptional": true,
-      "license": "MIT",
-      "peer": true
+      "license": "MIT"
     },
     "node_modules/@testing-library/dom": {
       "version": "10.4.0",
@@ -3088,7 +3081,8 @@
       "resolved": "https://registry.npmjs.org/@types/aria-query/-/aria-query-5.0.4.tgz",
       "integrity": "sha512-rfT93uj5s0PRL7EzccGMs3brplhcrghnDoV26NqKhCAS1hVo+WdNsPvE/yb6ilfr5hi2MEk6d5EWJTKdxg8jVw==",
       "dev": true,
-      "license": "MIT"
+      "license": "MIT",
+      "peer": true
     },
     "node_modules/@types/babel__core": {
       "version": "7.20.5",
@@ -3254,7 +3248,6 @@
       "integrity": "sha512-6mDvHUFSjyT2B2yeNx2nUgMxh9LtOWvkhIU3uePn2I2oyNymUAX1NIsdgviM4CH+JSrp2D2hsMvJOkxY+0wNRA==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "csstype": "^3.0.2"
       }
@@ -3265,7 +3258,6 @@
       "integrity": "sha512-9KQPoO6mZCi7jcIStSnlOWn2nEF3mNmyr3rIAsGnAbQKYbRLyqmeSc39EVgtxXVia+LMT8j3knZLAZAh+xLmrw==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "peerDependencies": {
         "@types/react": "^19.2.0"
       }
@@ -3355,7 +3347,6 @@
       "integrity": "sha512-tK3GPFWbirvNgsNKto+UmB/cRtn6TZfyw0D6IKrW55n6Vbs7KJoZtI//kpTKzE/DUmmnAFD8/Ca46s7Obs92/w==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@typescript-eslint/scope-manager": "8.46.4",
         "@typescript-eslint/types": "8.46.4",
@@ -3747,7 +3738,6 @@
       "integrity": "sha512-NZyJarBfL7nWwIq+FDL6Zp/yHEhePMNnnJ0y3qfieCrmNvYct8uvtiV41UvlSe6apAfk0fY1FbWx+NwfmpvtTg==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "bin": {
         "acorn": "bin/acorn"
       },
@@ -3781,7 +3771,6 @@
       "integrity": "sha512-B/gBuNg5SiMTrPkC+A2+cW0RszwxYmn6VYxB/inlBStS5nx6xHIt/ehKRhIMhqusl7a8LjQoZnjCs5vhwxOQ1g==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "fast-deep-equal": "^3.1.3",
         "fast-uri": "^3.0.1",
@@ -3890,6 +3879,7 @@
       "integrity": "sha512-b0P0sZPKtyu8HkeRAfCq0IfURZK+SuwMjY1UXGBU27wpAiTwQAIlq56IbIO+ytk/JjS1fMR14ee5WBBfKi5J6A==",
       "dev": true,
       "license": "Apache-2.0",
+      "peer": true,
       "dependencies": {
         "dequal": "^2.0.3"
       }
@@ -4184,7 +4174,6 @@
         }
       ],
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "caniuse-lite": "^1.0.30001726",
         "electron-to-chromium": "^1.5.173",
@@ -4820,7 +4809,6 @@
       "resolved": "https://registry.npmjs.org/convex/-/convex-1.29.0.tgz",
       "integrity": "sha512-uoIPXRKIp2eLCkkR9WJ2vc9NtgQtx8Pml59WPUahwbrd5EuW2WLI/cf2E7XrUzOSifdQC3kJZepisk4wJNTJaA==",
       "license": "Apache-2.0",
-      "peer": true,
       "dependencies": {
         "esbuild": "0.25.4",
         "prettier": "^3.0.0"
@@ -5139,6 +5127,7 @@
       "integrity": "sha512-0je+qPKHEMohvfRTCEo3CrPG6cAzAYgmzKyxRiYSSDkS6eGJdyVJm7WaYA5ECaAD9wLB2T4EEeymA5aFVcYXCA==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "engines": {
         "node": ">=6"
       }
@@ -5161,7 +5150,8 @@
       "resolved": "https://registry.npmjs.org/dom-accessibility-api/-/dom-accessibility-api-0.5.16.tgz",
       "integrity": "sha512-X7BJ2yElsnOJ30pZF4uIIDfBEVgF4XEBxL9Bxhy6dnrm5hkzqmsWHGTiHqRiITNhMyFLyAiWndIJP7Z1NTteDg==",
       "dev": true,
-      "license": "MIT"
+      "license": "MIT",
+      "peer": true
     },
     "node_modules/dompurify": {
       "version": "3.2.6",
@@ -5515,7 +5505,6 @@
       "integrity": "sha512-BhHmn2yNOFA9H9JmmIVKJmd288g9hrVRDkdoIgRCRuSySRUHH7r/DI6aAXW9T1WwUuY3DFgrcaqB+deURBLR5g==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@eslint-community/eslint-utils": "^4.8.0",
         "@eslint-community/regexpp": "^4.12.1",
@@ -6589,7 +6578,6 @@
       "resolved": "https://registry.npmjs.org/hono/-/hono-4.10.4.tgz",
       "integrity": "sha512-YG/fo7zlU3KwrBL5vDpWKisLYiM+nVstBQqfr7gCPbSYURnNEP9BDxEMz8KfsDR9JX0lJWDRNc6nXX31v7ZEyg==",
       "license": "MIT",
-      "peer": true,
       "engines": {
         "node": ">=16.9.0"
       }
@@ -7405,7 +7393,6 @@
       "integrity": "sha512-Cvc9WUhxSMEo4McES3P7oK3QaXldCfNWp7pl2NNeiIFlCoLr3kfq9kb1fxftiwk1FLV7CvpvDfonxtzUDeSOPg==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "cssstyle": "^4.2.1",
         "data-urls": "^5.0.0",
@@ -7678,6 +7665,7 @@
       "integrity": "sha512-h5bgJWpxJNswbU7qCrV0tIKQCaS3blPDrqKWx+QxzuzL1zGUzij9XCWLrSLsJPu5t+eWA/ycetzYAO5IOMcWAQ==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "bin": {
         "lz-string": "bin/bin.js"
       }
@@ -7733,7 +7721,6 @@
       "integrity": "sha512-jcByLnIFkd5gSXZmjNvS1TlmRhCXZjIzHYlaGkPlLIekG55JDR2Z4va9tZwCiP+/RDERiNhMOFu01xd6O5ct1Q==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "bin": {
         "marked": "bin/marked.js"
       },
@@ -7894,7 +7881,6 @@
       "integrity": "sha512-aChaVU/DO5aRPmk1GX8L+whocagUUpBQqoPtJk+cm7UOXUk87J4PeWCh6nNmTTIfEhiR9DI/+FnA8dln/hTK7g==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "funding": {
         "type": "opencollective",
         "url": "https://opencollective.com/mobx"
@@ -8817,6 +8803,7 @@
       "integrity": "sha512-Qb1gy5OrP5+zDf2Bvnzdl3jsTf1qXVMazbvCoKhtKqVs4/YK4ozX4gKQJJVyNe+cajNPn0KoC0MC3FUmaHWEmQ==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "ansi-regex": "^5.0.1",
         "ansi-styles": "^5.0.0",
@@ -8832,6 +8819,7 @@
       "integrity": "sha512-quJQXlTSUGL2LH9SUXo8VwsY4soanhgo6LNSm84E1LBcE8s3O0wpdiRzyR9z/ZZJMlMWv37qOOb9pdJlMUEKFQ==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "engines": {
         "node": ">=8"
       }
@@ -8842,6 +8830,7 @@
       "integrity": "sha512-Cxwpt2SfTzTtXcfOlzGEee8O+c+MmUgGrNiBcXnuWxuFJHe6a5Hz7qwhwe5OgaSYI0IJvkLqWX1ASG+cJOkEiA==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "engines": {
         "node": ">=10"
       },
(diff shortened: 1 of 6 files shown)


ksinghal:
Huge credit to @panzacoder for his implementation in Zodvex, which this is heavily inspired by!!

packages/convex-helpers/server/zod4.ts
```diff
@@ -0,0 +1,642 @@
...
+  query: QueryBuilder<DataModel, Visibility>,
+  customization: Customization<
+    GenericQueryCtx<DataModel>,
+    CustomArgsValidator,
+    CustomCtx,
+    CustomMadeArgs,
+    ExtraArgs
+  >,
+) {
+  return customFnBuilder(query, customization) as any;
```
natedunn: Casting `any` here would just override the types, no? Same for `zCustomMuation` and `zCustomAction`. Seems to be the case so far in my limited testing: 

Image

natedunn: Different spot but same idea? 

CleanShot 2025-10-16 at 11  49 33

natedunn: Very quick update before bed: re-adding `CustomBuilder` as seen in the v3 version, instead of `any` does seem to fix the issue related to the custom functions. We might need to generally review the `any` usage in order to preserve type-safety. I might have some time tomorrow to take a full pass on it.

natedunn: Didn't have enough time to complete a full pass yet. But making some progress though my own adjustments and using a bit from @panzacoder's wonderful zodvex project, mostly helping me with the `zodToConvex()` feature. Hopefully that is ok, let me know if you have any objection.

packages/convex-helpers/server/zod4.ts
```diff
@@ -0,0 +1,642 @@
+import { z } from "zod";
```
Nicolapps:
```suggestion
import { z } from "zod/v4/core";
```

I think that this is change is necessary. With the current import, users of convex-helpers that use Zod 3 would get the `/zod4` types incorrectly defined based on Zod 3 (see https://zod.dev/library-authors?id=how-to-support-zod-4). According to the same page, we need to import from `zod/v4/core`, unless we do something that can’t work with Zod Mini (but they strongly recommend against doing so).

packages/convex-helpers/server/zod4.ts
```diff
@@ -0,0 +1,642 @@
...
+  if (visited.has(zodValidator)) return v.any();
+  visited.add(zodValidator);
+
+  if (isZid(zodValidator)) {
+    const meta = registryHelpers.getMetadata(zodValidator);
+    return v.id((meta?.tableName as string) ?? "unknown");
+  }
+
+  if (zodValidator instanceof z.ZodDefault) {
+    const inner = zodValidator.unwrap() as unknown as z.ZodTypeAny;
```
Nicolapps: This doesn’t typecheck:

> Property 'unwrap' does not exist on type 'ZodDefault<any>'

packages/convex-helpers/server/zod4.ts
```diff
@@ -0,0 +1,642 @@
...
+  const base = z
+    .string()
+    .refine((s) => typeof s === "string" && s.length > 0, {
+      message: `Invalid ID for table "${tableName}"`,
+    })
+    .transform((s) => s as GenericId<TableName>)
+    .brand(`ConvexId_${tableName}`)
+    .describe(`convexId:${tableName}`);
+  registryHelpers.setMetadata(base, { isConvexId: true, tableName });
+  return base as z.ZodType<GenericId<TableName>>;
```
Nicolapps: This doesn’t typecheck:

```
Conversion of type 'ZodBranded<ZodEffects<ZodEffects<ZodString, string, string>, Id<TableName>, string>, `ConvexId_${TableName}`>' to type 'ZodType<Id<TableName>, ZodTypeDef, Id<TableName>>' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
  Types of property '_input' are incompatible.
    Type 'string' is not comparable to type 'Id<TableName>'.
      Type 'string' is not comparable to type '{ __tableName: TableName; }'.ts(2352)
```

packages/convex-helpers/server/zod4.ts
Nicolapps: Whoops, this was caused by my local installation using Zod 3 instead of Zod 4. I think that this issue can be fixed by importing from `zod/v4/core` (see my comment above).

packages/convex-helpers/server/zod4.ts
Nicolapps: (see above)

packages/convex-helpers/server/zod4.ts
```diff
@@ -0,0 +1,642 @@
...
+    CustomArgsValidator,
+    CustomCtx,
+    CustomMadeArgs,
+    ExtraArgs
+  >,
+) {
+  return customFnBuilder(action, customization) as any;
+}
+
+export function zBrand<T extends z.ZodTypeAny, B extends string | number | symbol>(
```
Nicolapps: Can you please document this export?

```suggestion
/**
 * Add a brand to a zod validator. Used like `zBrand(z.string(), "MyBrand")`.
 * Compared to zod's `.brand`, this also brands the input type, so if you use
 * the branded validator as an argument to a function, the input type will also
 * be branded. The normal `.brand` only brands the output type, so only the type
 * returned by validation would be branded.
 *
 * @param validator A zod validator - generally a string, number, or bigint
 * @param brand A string, number, or symbol to brand the validator with
 * @returns A zod validator that brands both the input and output types.
 */
export function zBrand<T extends z.ZodTypeAny, B extends string | number | symbol>(
```

packages/convex-helpers/server/zod4.ts
```diff
@@ -0,0 +1,642 @@
...
+    .describe(`convexId:${tableName}`);
+  registryHelpers.setMetadata(base, { isConvexId: true, tableName });
+  return base as z.ZodType<GenericId<TableName>>;
+}
+
+export const withSystemFields = <
+  Table extends string,
+  T extends { [key: string]: z.ZodTypeAny },
+>(tableName: Table, zObject: T) => {
+  return { ...zObject, _id: zid(tableName as any), _creationTime: z.number() } as const;
```
Nicolapps: I think that `as any` here can be avoided
```suggestion
  return { ...zObject, _id: zid(tableName), _creationTime: z.number() } as const;
```

packages/convex-helpers/server/zod4.ts
```diff
@@ -0,0 +1,642 @@
...
+};
+
+function isZid(schema: z.ZodTypeAny): boolean {
+  const m = registryHelpers.getMetadata(schema);
+  return !!(m && m.isConvexId && typeof m.tableName === "string");
+}
+
+function makeUnion(members: GenericValidator[]) {
+  const arr = members.filter((m): m is GenericValidator => !!m);
+  if (arr.length === 0) return v.any();
```
Nicolapps: This is incorrect: `v.any()` accepts all values while an empty union is a bottom type (it accepts nothing).

packages/convex-helpers/server/zod4.ts
```diff
@@ -0,0 +1,642 @@
...
+      out[k] = zodToConvexInternal(val as z.ZodTypeAny, visited);
+    }
+    return v.object(out);
+  }
+
+  if (zodValidator instanceof z.ZodUnion) {
+    const options = (zodValidator as any).options as z.ZodTypeAny[];
+    if (Array.isArray(options) && options.length > 0) {
+      const members = options.map((opt) => zodToConvexInternal(opt, visited));
+      return (makeUnion(members as [GenericValidator, GenericValidator, ...GenericValidator[]]) as unknown) as GenericValidator;
```
Nicolapps: Why are the casts necessary here?
```suggestion
      return makeUnion(members);
```

packages/convex-helpers/server/zod4.ts
```diff
@@ -0,0 +1,642 @@
...
+function makeUnion(members: GenericValidator[]) {
+  const arr = members.filter((m): m is GenericValidator => !!m);
+  if (arr.length === 0) return v.any();
+  if (arr.length === 1) return arr[0];
+  const [a, b, ...rest] = arr as [GenericValidator, GenericValidator, ...GenericValidator[]];
+  return v.union(a, b, ...rest);
+}
+
+// Narrow helpers for Convex validators
+function asValidator(x: unknown): any {
```
Nicolapps: Why is this necessary? By glancing at the usages of this function, it looks like the argument is already a `GenericValidator` at a place where something extending `GenericValidator` is expected

packages/convex-helpers/server/zod4.ts
```diff
@@ -0,0 +1,642 @@
...
+    return zodToConvexInternal((zodValidator as any).innerType as z.ZodTypeAny, visited);
+  }
+  if (zodValidator instanceof z.ZodTransform) {
+    const inner = (zodValidator as any).def?.schema as z.ZodTypeAny | undefined;
+    return inner ? zodToConvexInternal(inner, visited) : v.any();
+  }
+  return v.any();
+}
+
+export function zodToConvex<Z extends z.ZodTypeAny | ZodValidator>(zodSchema: Z) {
```
Nicolapps: Why is this function also accepting records? The homologous function from our Zed 3 support only accepts Zod 3 types, and `zodToConvexFields` is used to support objects of Zod types.

packages/convex-helpers/server/zod4.ts
```diff
@@ -0,0 +1,642 @@
...
+function asValidator(x: unknown): any {
+  return x as unknown as any;
+}
+
+// Zod v4 -> Convex (input)
+function zodToConvexInternal(
+  zodValidator: z.ZodTypeAny,
+  visited: Set<z.ZodTypeAny> = new Set(),
+): GenericValidator {
+  if (!zodValidator) return v.any();
```
Nicolapps: Why do this expect `zodValidator` to be possibly falsy here?

packages/convex-helpers/server/zod4.ts
```diff
@@ -0,0 +1,642 @@
...
+  if (visited.has(zodValidator)) return v.any();
```
Nicolapps: What is the goal of this?

It’s true that Zod now supports [recursive objects](https://zod.dev/v4?id=recursive-objects). However, there is no equivalent thing with Convex validators at the moment. I think that we should throw an error message instead of this happens.

packages/convex-helpers/server/zod4.ts
```diff
@@ -0,0 +1,642 @@
...
+}
+
+// Narrow helpers for Convex validators
+function asValidator(x: unknown): any {
+  return x as unknown as any;
+}
+
+// Zod v4 -> Convex (input)
+function zodToConvexInternal(
+  zodValidator: z.ZodTypeAny,
```
Nicolapps: `z.ZodTypeAny` is deprecated:

> @deprecated — Use z.ZodType (without generics) instead.

More importantly, don’t we want this function to take a generic argument instead? The “illegal” type casts that are being done below might be unsafe.

packages/convex-helpers/server/zod4.ts
```diff
@@ -0,0 +1,642 @@
...
+  }
+
+  if (zodValidator instanceof z.ZodString) return v.string();
+  if (zodValidator instanceof z.ZodNumber) return v.float64();
+  if (zodValidator instanceof z.ZodBigInt) return v.int64();
+  if (zodValidator instanceof z.ZodBoolean) return v.boolean();
+  if (zodValidator instanceof z.ZodNull) return v.null();
+  if (zodValidator instanceof z.ZodAny) return v.any();
+  if (zodValidator instanceof z.ZodUnknown) return v.any();
+  if (zodValidator instanceof z.ZodDate) return v.float64();
```
Nicolapps: This is incorrect, Convex’s `v.float64` doesn’t accept `Date` objects.

Maybe we want to give a default handling of dates with a codec instead?

natedunn:
@ksinghal I have big overhaul to fix some issues I found in type-safety and that were addressed by @Nicolapps. It also introduces some further fixes provided by @panzacoder's Zodvex. Would you mind if I made some edits to this PR or should I create a fresh one since its a fairly sizable edit? I am ok either way, I just need to get v4 support shipped finally. 😅

ksinghal:
please go ahead and make the edits here if you like!

panzacoder:
> @ksinghal I have big overhaul to fix some issues I found in type-safety and that were addressed by @Nicolapps. It also introduces some further fixes provided by @panzacoder's Zodvex. Would you mind if I made some edits to this PR or should I create a fresh one since its a fairly sizable edit? I am ok either way, I just need to get v4 support shipped finally. 😅

I'm also happy to contribute here. Don't mind that my code is being borrowed (obv it's OSS) but I probably would have opened a PR here already if I thought that was what the convex team wanted. 

@natedunn  if you've already put some work into it I'll wait for your edits, otherwise I can make some contributions later today / this week.

packages/convex-helpers/server/zod4.ts
```diff
@@ -0,0 +1,642 @@
...
+    return v.any();
+  }
+  if (zodValidator instanceof z.ZodLiteral) {
+    const val = (zodValidator as any).value as string | number | boolean | bigint | null;
+    return v.literal(val as any);
+  }
+  if (zodValidator instanceof z.ZodEnum) {
+    const options = (zodValidator as any).options as unknown[];
+    const literals = options.map((o) => v.literal(o as any)) as unknown as GenericValidator[];
+    if (literals.length === 0) return v.any();
```
Nicolapps: I think it’s more correct for the empty union to be a never type, not `any`

```suggestion
```

packages/convex-helpers/server/zod4.ts
```diff
@@ -0,0 +1,642 @@
...
+  return { ...zObject, _id: zid(tableName as any), _creationTime: z.number() } as const;
+};
+
+function isZid(schema: z.ZodTypeAny): boolean {
+  const m = registryHelpers.getMetadata(schema);
+  return !!(m && m.isConvexId && typeof m.tableName === "string");
+}
+
+function makeUnion(members: GenericValidator[]) {
+  const arr = members.filter((m): m is GenericValidator => !!m);
```
Nicolapps: Why do you need to check whether the elements are truthy? If a member is not truthy, it means that the function is called with incorrect types.

packages/convex-helpers/server/zod4.ts
```diff
@@ -0,0 +1,642 @@
...
+  if (arr.length === 0) return v.any();
+  if (arr.length === 1) return arr[0];
+  const [a, b, ...rest] = arr as [GenericValidator, GenericValidator, ...GenericValidator[]];
+  return v.union(a, b, ...rest);
```
Nicolapps: I think this can be simplified. In particular an empty union must be `v.union()` (bottom type), not `v.any()`. Also I don’t think it’s worth it to special case one-element unions.

```suggestion
  return v.union(...arr);
```

packages/convex-helpers/server/zod4.ts
```diff
@@ -0,0 +1,642 @@
...
+    return v.literal(val as any);
+  }
+  if (zodValidator instanceof z.ZodEnum) {
+    const options = (zodValidator as any).options as unknown[];
+    const literals = options.map((o) => v.literal(o as any)) as unknown as GenericValidator[];
+    if (literals.length === 0) return v.any();
+    return makeUnion(literals) as unknown as GenericValidator;
+  }
+  if (zodValidator instanceof z.ZodRecord) {
+    const valueType = (zodValidator as any).valueType as z.ZodTypeAny | undefined;
```
Nicolapps: Why do we expect `valueType` to be undefined here?

packages/convex-helpers/server/zod4.ts
```diff
@@ -0,0 +1,642 @@
...
+}
+
+export function zodOutputToConvexFields<Z extends ZodValidator>(zodShape: Z) {
+  const out: Record<string, GenericValidator> = {};
+  for (const [k, v_] of Object.entries(zodShape)) out[k] = zodOutputToConvexInternal(v_);
+  return out as { [k in keyof Z]: GenericValidator };
+}
+
+// Convex -> Zod minimal mapping (for tests)
+export function convexToZod(validator: GenericValidator): z.ZodTypeAny {
```
Nicolapps: Why is `convexToZod` here marked as being “for tests”? If it serves the same purposes as the equivalent function in v3, it can be useful for non-test purposes (e.g. integrating Convex validators with libraries expecting Zod validators, e.g. the AI SDK)

packages/convex-helpers/server/zod4.ts
```diff
@@ -0,0 +1,642 @@
...
+    const out: Record<string, GenericValidator> = {};
+    for (const [k, val] of Object.entries(shape)) {
+      out[k] = zodToConvexInternal(val as z.ZodTypeAny, visited);
+    }
+    return v.object(out);
+  }
+
+  if (zodValidator instanceof z.ZodUnion) {
+    const options = (zodValidator as any).options as z.ZodTypeAny[];
+    if (Array.isArray(options) && options.length > 0) {
```
Nicolapps: Why would `options` not be an array here?

Also an empty union should be a `never` type, i.e. `v.union()` (and not `v.any()`)

packages/convex-helpers/server/zod4.ts
```diff
@@ -0,0 +1,642 @@
...
+type ReturnValueInput<ReturnsValidator extends z.ZodTypeAny | ZodValidator | void> = [
+  ReturnsValidator,
+] extends [z.ZodTypeAny]
+  ? Returns<z.input<ReturnsValidator>>
+  : [ReturnsValidator] extends [ZodValidator]
+  ? Returns<z.input<z.ZodObject<ReturnsValidator>>>
+  : any;
+
+// (unused) type kept for reference in v3 version
+// type ReturnValueOutput<...> omitted in zod4 to reduce type depth
```
Nicolapps: As @natedunn indicated I think that this type is necessary if we want for the functions to have the right return value types without needing an explicit `returns` validator

natedunn:
@panzacoder Honestly my approach was just to edit this PR fairly minimally, but I ran into a few type issues that your approach seemed to solve already. Both of you guys had solid, sizable contributions. I'm just bringing them together and patching up a few things. 🙏

Was going to push changes up today but noticed that this PR isn't set up to accept edits. Mind checking that @ksinghal? Appreciate it

ksinghal:
Unbelievable - it seems I am unable to access the "allow maintainers to edit" box based on this discussion: https://github.com/orgs/community/discussions/5634. Can't believe that's the case so let me know if im just not looking in the right place.

natedunn:
@ksinghal Yeah, I was curious about how that flow would even work. Given its a few years old and there doesn't seem to be a work around, I think I should create another PR? Why I hate that is it robs you of the PR but also still doesn't allow you or @panzacoder to submit edits. Only reviews. 🙄

I could also just send you my updates through my own fork too. But again, no flow for further edits. Annoying.

panzacoder:
You could open a PR into this branch if you want to keep it focused here I think?

ksinghal:
Don't worry I don't care for the credit at all, I just want to help get it working. I am happy with whatever solution works for you!

natedunn:
I will submit a PR to your fork. That should work. I'm currently flying so I'll get to that as soon as I have a few minutes free.

Nicolapps:
Hi everyone, thanks a lot for your contributions! I reviewed this PR and there are a few things I think we should fix before adding Zod 4 support to convex-helpers.

The most important change is that I think we should handle dates differently. This PR tries to automatically convert dates to/from `v.float64()`, which I think is a bad idea. When converting values automatically, it makes the behavior more complex to users. In particular, it’s not clear how `z.union([z.date(), z.number()])` would be handled. This also requires users to use `toConvexJS`/`fromConvexJS` on the client, which doesn’t feel evident. Instead, Zod 4.1’s codecs API allows users to explicitly define how encoding/decoding should be done, which is ideal in this case. I wrote a design doc about how we can let users take advantage of codecs here: https://www.notion.so/convex-dev/Convex-Zod-4-Codecs-integration-294b57ff32ab809fa067fe63c740ca74?source=copy_link

A few other things on my mind:
- We should target `"zod": "^3.25.0 || ^4.0.0"` in our peer dependencies as suggested by https://zod.dev/library-authors, which will make sure that we can target both Zod 3 and Zod 4 at the same time. And inside of `zod4.ts`, we should import from `"zod/v4/core"`, as recommended by the same page. You can find more details about how we can handle Zod 3 + 4 in convex-helpers in [my design doc](https://www.notion.so/convex-dev/Convex-Zod-4-Codecs-integration-294b57ff32ab809fa067fe63c740ca74?source=copy_link).
- I think that the implementation of `zid` is more complex than it could be. In particular, I don’t think that we should use Zod branded values here, since Convex already uses its own branded string type for IDs. Using a Zod branded value would meant that the output type of Zod woudn’t be compatible with the `Id` type from Convex, which sounds incorrect.
- What is the goal of checking for visited types in `zodToConvexInternal`? Is it to support self-referential types, as in https://zod.dev/api?id=recursive-objects ? If so, it could be useful to add a comment + a test
- I found some things that I believe are bugs in this PR’s implementation:
  - Why is the code sometimes trying to remove falsy elements from arrays (e.g. `members.filter((m): m is GenericValidator => !!m)`)? This looks suspicious, because this is done on arrays where the elements should always be truthy.
  - An empty union type means that the value can never exist (`z.union([])` is equivalent to `never` in TypeScript). So `z.union([])` should be converted to `v.union()`, not `v.any()`.
  - When evaluating `z.ZodLazy`, I think that errors thrown by the schema should propagate, not be caught and ignored.
  - The handling of records looks incorrect: we shouldn’t assume that the key is `v.string()`, because it could also be a union of litterals. Also, in Zod 4, we must handle partial records differently.

I would be happy to write a PR with these changes. Thanks again for your help!

natedunn:
Hitting pause after skimming @Nicolapps comment and design. My PR addresses some, but not all of his concerns. Might be cleaner to let him PR it rather than trying to hit a target that's been moving a bit recently. 🤷‍♂️

ksinghal:
Thanks for the suggestions and input @natedunn ill accept your PR whenever its up and dive back in to help out at that point! Thanks to you as well @Nicolapps!

One note - you need to use zod 4.1 to use Codecs, which I think would be a clean way to implement bidirectional transforms like dates etc. Not sure if that affects the compatibility of zod v3

ksinghal:
The bugs you pointed out are valid issues, and zid implementation can definitely be simplified

natedunn:
@ksinghal I have PRed to your repo. @Nicolapps and gang feel free to reference or do your own thing. Thanks for all your help. Time to enjoy my vacation

Nicolapps:
I just merged #840 that brings full Zod 4 support to convex-helpers. Thanks for your help!

ianmacartney:
Anyone who wants to use it before we release it (likely Monday) can use `npm i convex-helpers@0.1.105-alpha.1` 🎉
Thanks everyone for your contributions & patience.

PR #824, opened 2025-10-21
Improve documentation of Zod

Nicolapps:


packages/convex-helpers/server/zod.ts
```diff
@@ -784,9 +803,62 @@ type ConvexValidatorFromZod<Z extends z.ZodTypeAny> =
...
+ *   for more details.
+ *
+ * Note that some values might be valid in Zod but not in Convex,
+ * in the same way that valid JavaScript values might not be valid
+ * Convex values for the corresponding Convex type.
+ * (see the limits of Convex data types on https://docs.convex.dev/database/types).
+ *
+ * #### Comparison with `zodOutputToConvex`
+ * | **zodToConvex**                                                                                        | {@link zodOutputToConvex}                                                                                             |
+ * | ------------------------------------------------------------------------------------------------------ | --------------------------------------------------------------------------------------------------------------------- |
```
ianmacartney: This table is so wide it's hard to read - let's try to make the docs readable with 80 character width display. likely means making this into a more concise table, inline the differences in one list, or something similar

Nicolapps: I fixed it in 4114d1628949e5951724d8d73ab8c5c249d9f6ed

ianmacartney:
only nit: the missing CLA test we should either update to not require for employees, or if you could add it to your descriptions then it won't be a red herring

Nicolapps:
@ianmacartney Did it here:

- https://github.com/get-convex/convex-helpers/pull/836

PR #839, opened 2025-11-05
Include tests in tsconfig

Nicolapps:


packages/convex-helpers/server/customFunctions.test.ts
```diff
@@ -62,7 +62,11 @@ const authenticatedQueryBuilder = customQuery(
 type AuthQueryCtx = CustomCtx<typeof authenticatedQueryBuilder>;
 
 // Example query that doesn't specify argument validation (no `args` param).
-export const unvalidatedArgsQuery = authenticatedQueryBuilder((ctx) => {
+export const unvalidatedArgsQuery: RegisteredQuery<
```
Nicolapps, line 65: This is to solve the following error:

```
The inferred type of 'addCU' cannot be named without a reference to '../../../node_modules/convex/dist/esm-types/server/registration.js'. This is likely not portable. A type annotation is necessary. ts(2742)
```

This error could also be solved by removing the `export`, but this is is not a problem since we’re in a test file.

package.json
```diff
@@ -27,7 +27,7 @@
   },
   "dependencies": {
     "classnames": "^2.3.2",
-    "convex": "^1.22.0",
+    "convex": "^1.28.2",
```
Nicolapps, line 30: For consistency with package.json in `packages/convex-helpers`

packages/convex-helpers/tsconfig.json
```diff
@@ -108,5 +108,5 @@
     "verbatimModuleSyntax": true
   },
   "include": ["."],
-  "exclude": ["node_modules", "dist", "**/*.test.ts"]
```
ianmacartney, line 111: we need this to avoid emitting .test files. maybe we override this when doing type checking

package.json
ianmacartney: we could make all of these pinned dev dependencies so renovate keeps them up to date? Since we don't actually need to declare regular dependencies?

PR #840, opened 2025-11-11
Zod 4 support

Nicolapps:
This pull request adds full support for Zod 4 in convex-helpers.

## Migration

The methods in `convex-helpers/server/zod` have been moved to `convex-helpers/server/zod3`. The new Zod 4-compatible methods can be imported from  `convex-helpers/server/zod4`.

For convenience, some methods in `convex-helpers/server/zod` (`zodToConvex`, `zodOutputToConvex`, `zodToConvexFields`, `zodOutputToConvexFields`, and `withSystemFields`) now accept both Zod 3 and Zod 4 schemas. If you’re using Zod 4 in your codebase, we strongly recommend that you import directly from `convex-helpers/server/zod4`, as it simplifies the work that the TypeScript compiler needs to do.

## Codecs support
When using Zod 4.1, developers can benefit from [Zod codecs](https://zod.dev/codecs) to roundtrip values to/from functions and the database. As explained in https://www.notion.so/convex-dev/Convex-Zod-4-Codecs-integration-294b57ff32ab809fa067fe63c740ca74?source=copy_link, the code of convex-helpers doesn’t need to do anything related to codecs:

- When using codecs to encode _function arguments_, developers can use codecs with `Input = Convex value`  and `Output = Developer-facing value`
- When using codecs to encode _function return values_, developers can use codecs with `Input = Developer-facing value` and `Output = Convex value`
- When using codecs to store functions in the database, developers can use codecs with `Input = Convex value` and `Output = Developer-facing value`. (convex-helpers doesn’t provide a database wrapper that encodes values with Zod, but if you’re writing one, you would want to call the `.encode()` and `.decode()` functions of Zod schemas).

This strategy allows developers to keep using `z.transform()` in both `args` and `returns`, while keeping support for Zod 3 and 4.0.

## Implementation details
* `zid` is implemented as a `z.custom` call with a registry that stores the table name in a `WeakMap`. I considered extending `z.ZodCustom` or `z.ZodType`, but this isn’t officially supported by Zod (see https://discord.com/channels/893487829802418277/1359688107594748054/1367295316365152308).
* `zodToConvex` and `zodOutputToConvex` support circular schemas (`v.any()` is returned when a type is used more than once).
* `z.templateLiteral()` is now supported.
* `z.tuple()` is now supported: it might return validators that are more complex than necessary (e.g. `[number, number]` will be converted to `v.union(v.number(), v.number())`).
* `z.literal()` now supports literals with multiple values (e.g. `z.literal([1, 2, 3])`) correctly.
* For some types, `z.tuple()` and `z.literal()`, the output type is loosened to not assume a specific order in the resulting `VUnion`. This is necessary since the type information in Zod only contains a union, which don’t have a guaranteed order in TypeScript. The Zod 3 implementation didn’t handle this correctly (but wasn’t updated).
* `z.record()` now uses `v.object()` when the key is a literal (or union of literals), since `v.union(v.literal(…), …)` isn’t a valid Convex validator.
* Methods that take a Zod schema support both Zod 4 and Zod 4 mini. Methods that return a Zod schema don’t support Zod Mini, but support could be added in the future if necesary.

## Acknowledgments
This pull request is inspired by the work done by @ksinghal in https://github.com/get-convex/convex-helpers/pull/818 and by @natedunn in https://github.com/vendpark/convex-helpers/pull/1 and https://github.com/vendpark/convex-helpers/pull/2. Thanks to both of you!

Co-Authored-By: Karan Singhal <ksinghal@users.noreply.github.com>
Co-Authored-By: Nate Dunn <hello@natedunn.net>

### DIFF
--- a/packages/convex-helpers/server/zod4.ts
+++ b/packages/convex-helpers/server/zod4.ts
@@ -1474,9 +1474,7 @@ type IsUnknownOrAny<T> =
     ? true
     : // unknown?
       unknown extends T
-      ? [T] extends [unknown]
-        ? true
-        : false
+      ? true
       : false;
 
 function zodToConvexCommon<Z extends zCore.$ZodType>(


packages/convex-helpers/server/zod4.ts
```diff
@@ -0,0 +1,1978 @@
...
+          ? [Expand<A & ObjectType<CustomArgsValidator>>]
+          : [ObjectType<CustomArgsValidator>]
+    >,
+    ReturnsZodValidator extends void
+      ? ReturnValue
+      : ReturnValueOutput<ReturnsZodValidator>
+  >;
+};
+
+function customFnBuilder(
```
ianmacartney, line 833: This function is almost identical to zod3 except for a few lines. Ok for now, but is the eventual plan to deprecate zod3 or to unify or keep as is? It's ok to keep it all separate. maybe a comment to keep them up to date is enough?

packages/convex-helpers/server/zod4.ts
```diff
@@ -0,0 +1,1978 @@
...
+  | zCore.$ZodUndefined
+  | zCore.$ZodVoid
+  ? true
+  : false;
+
+type IsUnion<T, U extends T = T> = T extends unknown
+  ? [U] extends [T]
+    ? false
+    : true
+  : false;
```
ianmacartney, line 1447: This is some more magic I don't yet understand.

packages/convex-helpers/server/zod4.ts
```diff
@@ -0,0 +1,1978 @@
...
+type IsUnknownOrAny<T> =
+  // any?
+  0 extends 1 & T
+    ? true
+    : // unknown?
+      unknown extends T
+      ? [T] extends [unknown]
+        ? true
+        : false
+      : false;
```
ianmacartney, line 1478: are these standard type utilities, llm-produced, or your own invention?

packages/convex-helpers/server/zod4.ts
```diff
@@ -0,0 +1,1978 @@
...
+      }
+      return v.object(fields);
+    }
+
+    return v.record(
+      isValidRecordKey(keyValidator) ? keyValidator : v.string(),
+      vRequired(valueValidator),
+    );
+
+    function extractStringLiterals(
```
ianmacartney: any reason these helper functions can't be outside of the function body? not sure if it has an impact/ the interpreter is smart enough, but I wouldn't want every invocation of this function to create these functions if not necessary

packages/convex-helpers/server/zod4.ts
```diff
@@ -0,0 +1,1978 @@
...
+                              >
+                            ? VUnion<
+                                NotUndefined<Type>,
+                                Members,
+                                "required",
+                                FieldPaths
+                              >
+                            : never;
+
+function vRequired(validator: GenericValidator) {
```
ianmacartney, line 1940: one optimization would be to return the existing validator if it's already required

packages/convex-helpers/server/zod4.ts
```diff
@@ -0,0 +1,1978 @@
...
+/** Stores the table names for each `Zid` instance that is created. */
+const _zidRegistry = zCore.registry<{ tableName: string }>();
+
+// #endregion
+
+// #region Implementation: Utilities
+
+type NotUndefined<T> = Exclude<T, undefined>;
+
+type VRequired<T extends Validator<any, OptionalProperty, any>> =
```
ianmacartney, line 1875: do you think this would be useful in the core package? or just inline it here to start?

packages/convex-helpers/server/zod4.ts
Nicolapps: This comes from https://stackoverflow.com/a/59687759/4652564. I don’t understand in detail how it works, but from my testing it looks like it works correctly.

packages/convex-helpers/server/zod4.ts
Nicolapps: Thanks, I added a comment in fdf6e5b23e70520cecd7f59e76fa172e1976612f

packages/convex-helpers/server/zod4.ts
Nicolapps: No, they were just here to define them in the only place where they are used. I moved them to the global scope in e43940ddcc6832c25c6853a3a86e006d8c72a2d6

packages/convex-helpers/server/zod4.ts
Nicolapps: Thanks, I did it in 54e0622725e77f7de84673b4bc07f7edc84f5e76

packages/convex-helpers/server/zod4.ts
Nicolapps: Yes, this is a good idea. I can do this as a follow-up task, I created an issue at https://github.com/get-convex/convex-helpers/issues/847

packages/convex-helpers/server/zod4.ts
Nicolapps: For this one I used a LLM to solve the following problems:

- knowing whether a generic type is `any`
- knowing whether a generic type is `unknown`

I got consistent answers after asking multiple times and my type tests pass, so I think that the solution is correct

Nicolapps: I see the same solution for `IsAny` being used here: https://github.com/remix-run/react-router/blob/main/packages/react-router/lib/types/utils.ts

and I also see a simpler `IsUnknown` type in the TypeScript codebase:

https://github.com/microsoft/TypeScript/blob/cc2610fd347127c28609b4fca6fa9d55ee5063c6/tests/cases/conformance/types/unknown/unknownType2.ts#L3

so I simplified the type in a262a7461140187229f188564a9bf699f38f66ee

Nicolapps (commented):
Thanks @ianmacartney for your review!

PR #846, opened 2025-11-14
Run `npm run format`

Nicolapps:
This seems to revert https://github.com/get-convex/convex-helpers/commit/b08dcc35a176956dab3d7bd18e93370fd30b76e2. I’m not sure why, but this was causing the CI in the Zod PR to fail.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,6 +19,7 @@ npm install
 Adding helpers usually involves:
 
 1. Adding code (and corresponding .test.ts file) to:
+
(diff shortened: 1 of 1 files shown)


ianmacartney:
yeah this is odd - I wonder if there's a different version of prettier installed in CI? I made the change b/c CI was failing due to the linting of CHANGELOG.md

Can you verify if you're using the latest prettier when formatting?

Nicolapps:
Thanks @ianmacartney for the fix!

PR #853, opened 2025-11-18
Make triggers work with the explicit table API

Nicolapps:


packages/convex-helpers/server/triggers.ts
```diff
@@ -168,43 +169,61 @@ export class DatabaseWriterWithTriggers<
...
   ): Promise<GenericId<TableName>> {
     return this.writer.insert(table, value);
   }
-  patch<TableName extends TableNamesInDataModel<DataModel>>(
-    id: GenericId<TableName>,
-    value: Partial<DocumentByName<DataModel, TableName>>,
-  ): Promise<void> {
-    return this.writer.patch(id, value);
+
+  patch(arg0: any, arg1: any, arg2?: any): Promise<void> {
```
ianmacartney, line 217: do these `any` types show up in the customer's app? Or are they sufficiently hidden by the library otherwise?

packages/convex-helpers/server/triggers.ts
```diff
@@ -219,6 +238,73 @@ export function writerWithTriggers<
...
   isWithinTrigger: boolean = false,
 ): GenericDatabaseWriter<DataModel> {
+  async function _patch<TableName extends TableNamesInDataModel<DataModel>>(
+    tableName: TableName | null,
+    id: GenericId<TableName>,
+    value: Partial<DocumentByName<DataModel, TableName>>,
+  ): Promise<void> {
+    if (!tableName) {
+      return await innerDb.patch(id, value);
+    }
```
ianmacartney, line 296: is _patch getting called from elsewhere? Or could we inline it? It wasn't clear why tableName might be null here without going to the (only I think?) call site. I'm partial to inlining these helper functions unless they're serving another purpose? Are they here to provide more type safety?

packages/convex-helpers/server/triggers.ts
Nicolapps: Thanks, I fiexd the types in 875fe205321e000b157c58c75623a146e81075aa

packages/convex-helpers/server/triggers.ts
Nicolapps: Yeah, the only reason why I’m doing that is to get the right types. Do you have a better idea?

packages/convex-helpers/server/triggers.ts
```diff
@@ -168,43 +169,99 @@ export class DatabaseWriterWithTriggers<
...
+  delete<TableName extends TableNamesInDataModel<DataModel>>(
+    table: NonUnion<TableName>,
+    id: GenericId<TableName>,
+  ): Promise<void>;
+  delete(id: GenericId<TableNamesInDataModel<DataModel>>): Promise<void>;
+  delete(arg0: any, arg1?: any): Promise<void> {
+    return arg1 !== undefined
+      ? this.writer.delete(
+          arg0,
+          // @ts-expect-error -- delete supports 2 args since convex@1.25.4, but the type is marked as internal
```
Nicolapps: I’ll be able to remove these @ts-expect-error comments whenever we start using 1.30.0 internally for building convex-helpers. I updated the Convex peer dependency of helpers to target ^1.25.4: all versions of `convex` support the variant with an explicit table name argument, even if the published types don’t advertise it.

Nicolapps:
> looking good so far - also remember to update streams & relationship helpers (and any other spots I'm forgetting)

@ianmacartney Thanks! I’m going to do a separate PR for row-level security. I don’t see custom implementations of `get`/`replace`/`patch`/`delete` in relationships/streams, is there something I’m missing?

packages/convex-helpers/server/triggers.ts
```diff
@@ -168,43 +169,99 @@ export class DatabaseWriterWithTriggers<
...
+  ): Promise<void>;
+  replace(arg0: any, arg1: any, arg2?: any): Promise<void> {
+    return arg2 !== undefined
+      ? this.writer.replace(
+          arg0,
+          arg1,
+          // @ts-expect-error -- replace supports 3 args since convex@1.25.4, but the type is marked as internal
+          arg2,
+        )
+      : this.writer.replace(arg0, arg1);
```
ianmacartney:
```suggestion
    return this.writer.replace(
      arg0,
      arg1,
      // @ts-expect-error -- replace supports 3 args since convex@1.25.4, but the type is marked as internal
      arg2,
    );
```

ianmacartney:
ah yeah stream only is on the queries, my mistake. And relationship helpers
I think we can wait until it ships and either add deprecated overloads or
deprecate some in general?

packages/convex-helpers/server/triggers.ts
Nicolapps: Thanks, I simplified it in 5d4237d6e04eb9ee403aa1e2978a19a4548fc1ef

Nicolapps:
> Looks good - is there some sanity check somewhere that it conforms to the DatabaseReader / Writer types? e.g. that ctx including a wrapped db will not have a type error? I guess the wrapDatabaseWriter controls that.

Yeah, since `writerWithTriggers` returns a `GenericDatabaseWriter` it is checked that way. The only part where type safety is worse is the methods that take `any`, for example:

```
patch: async (arg0: any, arg1: any, arg2?: any) => {
  const [tableName, id, value] =
    arg2 !== undefined
      ? [arg0, arg1, arg2]
      : [_tableNameFromId(innerDb, triggers.registered, arg0), arg0, arg1];
  return await _patch(tableName, id, value);
}
```

i.e. if we updated `patch` to take other arguments another day, it would still typecheck. I changed the implementation in 9adeba1c4e2421ffe1b2614186aa5dc1f756b5c6: by duplicating the type definition, we avoid this scenario. Let me know your thoughts on this!

----

> ah yeah stream only is on the queries, my mistake. And relationship helpers
> I think we can wait until it ships and either add deprecated overloads or
> deprecate some in general?

I see what you mean. We don’t need to change anything for now as both APIs are supported.

PR #855, opened 2025-11-20
Zod 4 support: fix empty function args

### LINKED ISSUE #854: Zod v3 to v4 `zCustomQuery` migration breaking `useQuery` empty parameters
When you migrated from:

```ts
import { zCustomQuery } from 'convex-helpers/server/zod';  // Zod v3
```

 To:

```ts
import { zCustomQuery } from 'convex-helpers/server/zod4'; // Zod v4
```

Functions defined without args now require `{}` to be passed in `useQuery`.

Take this example:

```ts
// convex/users.ts (updated to use Zod 4)

// import { zCustomAction, zCustomMutation, zCustomQuery } from 'convex-helpers/server/zod';
import { zCustomAction, zCustomMutation, zCustomQuery } from 'convex-helpers/server/zod4';

export const zQuery = zCustomQuery(query, ...);

export const getUser= zQuery({
  args: {},
  handler: async (ctx) => {
    //...
  },
});

```

```tsx
// ListUsers.tsx (frontend)

import { useQuery } from 'convex-helpers/react';

// works when using `zCustomQuery` from Zod v3, but not Zod4
const userQuery = useQuery(api.users.getUser);

// workaround for Zod4 is to add `{}` args
const userQuery = useQuery(api.users.getUser, {});
```

Nicolapps:
Fixes #854

### DIFF
--- a/packages/convex-helpers/server/zod4.ts
+++ b/packages/convex-helpers/server/zod4.ts
@@ -968,10 +968,11 @@ type ArgsInput<ArgsValidator extends ZodFields | zCore.$ZodObject<any> | void> =
   [ArgsValidator] extends [zCore.$ZodObject<any>]
     ? [zCore.input<ArgsValidator>]
     : ArgsValidator extends Record<string, never>
-      ? [{}]
+      ? // eslint-disable-next-line @typescript-eslint/no-empty-object-type
+        [{}]
       : [ArgsValidator] extends [Record<string, z.ZodTypeAny>]
         ? [zCore.input<zCore.$ZodObject<ArgsValidator, zCore.$strict>>]
-      : OneArgArray;
+        : OneArgArray;
 
 // The args after they've been validated: passed to the handler
 type ArgsOutput<


convex/zodFunctionsExample.ts
```diff
@@ -0,0 +1,36 @@
+import { zCustomQuery } from "convex-helpers/server/zod4";
+import { query } from "./_generated/server";
+import { NoOp } from "convex-helpers/server/customFunctions";
+import { z } from "zod/v4";
+
+export const zQuery = zCustomQuery(query, NoOp);
+
+export const noArgs = zQuery({
+  args: {},
+  handler: async (ctx) => {
```
ianmacartney:
```suggestion
  handler: async (_ctx) => {
```

packages/convex-helpers/server/zod4.functions.test.ts
```diff
@@ -258,6 +267,12 @@ describe("zCustomQuery, zCustomMutation, zCustomAction", () => {
       >();
     });
 
+    test("zCustomQuery with no args", async () => {
+      const t = convexTest(schema, modules);
+      const response = await t.query(testApi.testQueryNoArgs);
```
ianmacartney: another way we can test the behavior of it 
```suggestion
      const response = await t.run((ctx) => ctx.runQuery(testApi.testQueryNoArgs));
```
Since `ctx.runQuery` also expects no args to make args optional

Nicolapps: Thanks, I did it in c76ca213e2b22e4437e0672403da7da354325b27

PR #863, opened 2025-11-24
remove console log

gary-ix:
This error message clutters our logs. If just removing it is not okay I am open to brainstorming. (Like an optional param to disable it)


Nicolapps (approved):
Thank you for your contribution! The goal of this log message was to warn the user that by doing things like `z.string().transform(str => str.toUpperCase())`, the resulting Convex validator isn’t as strict (e.g. `z.any()` instead of `z.string()`), but using `transform` isn’t wrong in itself so we agree in can be removed.

PR #872, opened 2025-12-05
Zod 4 support: fix `VUnion<… …[]>` handling in convexToZod

### LINKED ISSUE #861: Incompatibility between literals helper and zod
Example

when generating a zod schema

```ts
import { convexToZod } from "convex-helpers/server/zod4";
export const PersonSchema = convexToZod(Person.table.validator);
```

We get a proper ZodValidator when we define the table as followed

```ts
export const Person= Table("persons", {
  firstName: v.string(),
  lastName: v.string(),
  gender: v.union(v.literal("male"), v.literal("female"), v.literal("other")),
});
```

But when we define it as

```ts
export const Person= Table("persons", {
  firstName: v.string(),
  lastName: v.string(),
   gender: literals("male", "female", "other"),
});
```

The zod PersonSchema gives a never type to the gender field

Nicolapps:
Fixes #861


ianmacartney (commented):
looks like it's only tests right now, but the tests look good, feel free to land once they're passing

packages/convex-helpers/server/zod4.convextozod.test.ts
```diff
@@ -205,6 +206,46 @@ describe("convexToZod", () => {
...
+    enum Gender {
+      Male = "male",
+      Female = "female",
+      Other = "other",
+    }
+
+    testConvexToZod(
+      v.union(...Object.values(Gender).map(v.literal)),
+      ignoreZodUnionOrder(
+        z.union([
```
Nicolapps, line 237: I need to add `ignoreZodUnionOrder` to the test, because `z.union([…])` stores the variants as an ordered tuple (`z.ZodUnion<readonly [z.ZodLiteral<Gender.Male>, z.ZodLiteral<Gender.Female>, z.ZodLiteral<Gender.Other>]>`). At the same time, `Object.values(Gender).map(v.literal)` is a `VLiteral<Gender, "required">[]`, so we can’t determine the right order from its type.

I added the `ignoreZodUnionOrder` helper to reduce the precision of the `z.union` type. This is similar to the `ignoreUnionOrder` helper we have in `zod4.zodtoconvex.test.ts`. This PR also adds a few tests for this helper, similarly to how we have tests for `ingoreUnionOrder` in `zod4.zodtoconvex.test.ts`.

packages/convex-helpers/server/zod4.convextozod.test.ts
```diff
@@ -205,6 +206,46 @@ describe("convexToZod", () => {
...
     });
   });
+
+  // https://github.com/get-convex/convex-helpers/issues/861
+  test("regression: literals helper", () => {
+    testConvexToZod(
+      v.object({
+        firstName: v.string(),
+        lastName: v.string(),
+        gender: literals("male", "female", "other"),
```
ianmacartney: nit: probably safer to pick a different enum, given how many different interpretations of gender there are. - maybe role: user|admin

packages/convex-helpers/server/zod4.convextozod.test.ts
```diff
@@ -227,3 +268,80 @@ function testConvexToZod<
...
+  Members extends readonly z.ZodLiteral<zCore.util.Literal>[],
+>(
+  union: z.ZodUnion<Members>,
+): Members extends readonly z.ZodLiteral<infer V>[]
+  ? // ↓ tuple of single-value literals → array of one multi-value literal
+    z.ZodUnion<readonly z.ZodLiteral<V>[]>
+  : never {
+  return union as unknown as Members extends readonly z.ZodLiteral<infer V>[]
+    ? z.ZodUnion<readonly z.ZodLiteral<V>[]>
+    : never;
```
ianmacartney: if you're casting here, then is the return type useful? maybe we could just have one copy of this transformation?

ianmacartney (approved):
looks good, thanks for digging into it!

packages/convex-helpers/server/zod4.convextozod.test.ts
Nicolapps: Thanks, done in 4bb329a

packages/convex-helpers/server/zod4.convextozod.test.ts
Nicolapps: Thanks, I simplified the implementation in 75288bd

PR #879, opened 2025-12-15
fix: call convexToJSON on args being passed to JSON.stringify

RhysSullivan:
Causes a runtime crash without this

----

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved query memoization to properly detect changes in query parameters, ensuring queries re-run when needed and avoid unnecessary re-executions when parameters are structurally equivalent.


ianmacartney (approved):
Thanks!

PR #892, opened 2026-01-04
fix: guard against non-object values in WeakSet for Zod 4

### LINKED ISSUE #891: WeakSet Error in `convex-helpers/server/zod4` with Zod 4
## Summary

When using `convex-helpers/server/zod4` with Zod 4, Convex deployments fail during esbuild's static analysis phase with the error: **"Invalid value used in weak set"**. This occurs in the `zodToConvex()` and `zodOutputToConvex()` functions when they attempt to add non-object values to a `WeakSet`.

## Environment

- **convex-helpers**: `0.1.108`
- **zod**: `^4.3.4` (or `4.3.5`)
- **convex**: `^1.31.2`
- **Node.js**: (any version)
- **Build tool**: esbuild (via Convex CLI)

### Error Message
```
Failed to analyze mutations/conversations/addMessage.js: 
Uncaught TypeError: Invalid value used in weak set
    at add [as add] (<anonymous>)
    at map [as map] (<anonymous>)
```

### When It Occurs
- **Trigger**: Running `npx convex dev` or `npx convex deploy`
- **Location**: During esbuild's static analysis phase (not at runtime)
- **Affected**: ALL mutations and queries using `zCustomMutation`/`zCustomQuery` with Zod 4 schemas
- **Files**: `convex-helpers/server/zod4.ts` (lines 427-440 and 517-540)

## Root Cause

The issue occurs in the `zodToConvex()` and `zodOutputToConvex()` functions, which use a `WeakSet` to track visited Zod validators during schema conversion to prevent circular references:

```typescript
export function zodToConvex<Z extends zCore.$ZodType>(
  validator: Z,
): ConvexValidatorFromZod<Z, "required"> {
  const visited = new WeakSet<zCore.$ZodType>();

  function zodToConvexInner(validator: zCore.$ZodType): GenericValidator {
    // Circular validator definitions are not supported by Convex validators,
    // so we use v.any() when there is a cycle.
    if (visited.has(validator)) {
      return v.any();
    }
    visited.add(validator); // ❌ ERROR: Non-object value passed here
    // ...
  }
}
```

**The Problem**: During esbuild's static analysis, some values passed to `zodToConvexInner()` may not be objects (e.g., `null`, `undefined`, or primitives). `WeakSet.add()` only accepts objects, so attempting to add a non-object value throws the "Invalid value used in weak set" error.

**Why It Happens**: Zod 4's internal structure differs from Zod 3, and during static analysis, esbuild may encounter intermediate values or properties that aren't objects. The recursive nature of `zodToConvexInner()` means it can be called with values extracted from Zod's internal `_zod.def` structure, which may include non-object values.

## Minimal Reproduction Example

Here's a minimal example that triggers the error:

```typescript
// convex/mutations/example.ts
import { zCustomMutation } from "convex-helpers/server/zod4";
import { mutation } from "./_generated/server";
import { z } from "zod";

const zMutation = zCustomMutation(mutation, { input: async (ctx, args) => ({ ctx, args: {} }) });

export const example = zMutation({
  args: {
    conversationId: z.string(),
    message: z.string(),
  },
  handler: async (ctx, args) => {
    return { success: true };
  },
});
```

When running `npx convex dev`, this will fail with the WeakSet error during the static analysis phase.

### Why This Example Triggers the Error

The issue isn't because `z.string()` is a string primitive. It's about how the recursive processing encounters non-object values during esbuild's static analysis:

1. **Processing Flow**: When `zodToConvexFields()` processes the shape object, it calls `zodToConvex(z.string())` for each field
2. **Nested Property Access**: During static analysis, esbuild evaluates nested property accesses like `validator._zod.def.shape`, `validator._zod.def.innerType`, etc.
3. **Non-Object Values**: Some of these property accesses may yield `undefined`, `null`, or other non-object values during static evaluation
4. **Recursive Calls**: These non-object values get passed to `zodToConvexInner()` recursively, which then tries to add them to the WeakSet (which only accepts objects)

**Key Point**: Zod 4's changed internal structure (`_zod.def` vs Zod 3's structure) means that during static analysis, esbuild's evaluation of nested property accesses can produce non-object intermediate values that weren't present in Zod 3.

## Real-World Example

From our codebase:

```typescript
// src/convex/mutations/conversations/addMessage.ts
import { zAuthenticatedMutation } from "../../helpers/zodHelpers";
import { zid } from "convex-helpers/server/zod4";
import { ConvexAddMessageDTO } from "../../../lib/systems/ai-chat/dto/mutations/message.dto";

export const addMessage = zAuthenticatedMutation({
    args: ConvexAddMessageDTO.extend({
        conversationId: zid(TABLE_NAME.CONVERSATIONS)
    }),
    handler: async (ctx, args, userId) => {
        // ... handler implementation
    },
    returns: zid(TABLE_NAME.MESSAGES)
});
```

This mutation fails during deployment with the WeakSet error, even though the schema is valid and would work at runtime.

## Proposed Fix

Add a guard check to ensure the validator is an object before using WeakSet operations:

```typescript
export function zodToConvex<Z extends zCore.$ZodType>(
  validator: Z,
): ConvexValidatorFromZod<Z, "required"> {
  const visited = new WeakSet<zCore.$ZodType>();

  function zodToConvexInner(validator: zCore.$ZodType): GenericValidator {
    // Guard: Ensure validator is an object before using WeakSet
    if (typeof validator !== "object" || validator === null) {
      return v.any();
    }
    
    // Circular validator definitions are not supported by Convex validators,
    // so we use v.any() when there is a cycle.
    if (visited.has(validator)) {
      return v.any();
    }
    visited.add(validator);

    const result =
      validator instanceof zCore.$ZodDefault
        ? v.optional(zodToConvexInner(validator._zod.def.innerType))
        : validator instanceof zCore.$ZodPipe
          ? zodToConvexInner(validator._zod.def.in)
          : zodToConvexCommon(validator, zodToConvexInner);

    // After returning, we remove the validator from the visited set because
    // we only want to detect circular types, not cases where part of a type
    // is reused (e.g. `v.object({ field1: mySchema, field2: mySchema })`).
    visited.delete(validator);
    return result;
  }

  return zodToConvexInner(validator) as any;
}
```

Apply the same fix to `zodOutputToConvex()`:

```typescript
export function zodOutputToConvex<Z extends zCore.$ZodType>(
  validator: Z,
): ConvexValidatorFromZodOutput<Z, "required"> {
  const visited = new WeakSet<zCore.$ZodType>();

  function zodOutputToConvexInner(validator: zCore.$ZodType): GenericValidator {
    // Guard: Ensure validator is an object before using WeakSet
    if (typeof validator !== "object" || validator === null) {
      return v.any();
    }
    
    // Circular validator definitions are not supported by Convex validators,
    // so we use v.any() when there is a cycle.
    if (visited.has(validator)) {
      return v.any();
    }
    visited.add(validator);

    const result =
      validator instanceof zCore.$ZodDefault
        ? zodOutputToConvexInner(validator._zod.def.innerType)
        : validator instanceof zCore.$ZodPipe
          ? zodOutputToConvexInner(validator._zod.def.out)
          : validator instanceof zCore.$ZodTransform
            ? v.any()
            : zodToConvexCommon(validator, zodOutputToConvexInner);

    visited.delete(validator);
    return result;
  }

  return zodOutputToConvexInner(validator) as any;
}
```

## Why This Fix Works

1. **Prevents WeakSet errors**: The guard ensures only objects are added to the WeakSet
2. **Safe fallback**: Returning `v.any()` for non-object values is a reasonable fallback since these shouldn't occur in normal Zod schemas
3. **Preserves functionality**: Circular reference detection still works for valid object validators
4. **No breaking changes**: The fix is backward compatible and doesn't change the API

## Impact

- **Severity**: High - Blocks all Convex deployments using Zod 4 with `convex-helpers`
- **Workaround**: Currently, users must either:
  - Downgrade to Zod 3
  - Apply a local patch using `patch-package`
  - Avoid using `zCustomMutation`/`zCustomQuery` for affected functions

## Related Issues

This issue is specific to Zod 4. The same code works correctly with Zod 3 (`convex-helpers/server/zod3`), suggesting that Zod 4's internal structure changes are causing the problem during static analysis.

## Testing

To verify the fix:

1. Apply the guard checks to both functions
2. Run `npx convex dev` with a project using Zod 4
3. Verify that mutations/queries using `zCustomMutation`/`zCustomQuery` deploy successfully
4. Test that circular reference detection still works correctly

## Additional Notes

- This error only occurs during esbuild's static analysis, not at runtime
- The error is consistent and reproducible with any Zod 4 schema used in Convex functions
- We've verified that the issue is not caused by:
  - `.describe()` calls (with or without template literals)
  - `zid()` usage
  - `.extend()` method
  - Complex schemas
  - Wrapper functions

Borouch:
Fixes #891
Add type guard to prevent 'Invalid value used in weak set' error during esbuild static analysis. The WeakSet.add() method only accepts objects, but during static analysis, some values passed to zodToConvex and zodOutputToConvex may be primitives or null.

This fix adds a guard check before using WeakSet operations to ensure the validator is an object, returning v.any() as a fallback for non-object values.

Fixes the issue where mutations using zCustomMutation with Zod 4 schemas fail during npx convex dev/deploy with 'Invalid value used in weak set' error.

----

### DIFF
--- a/packages/convex-helpers/server/zod4.ts
+++ b/packages/convex-helpers/server/zod4.ts
@@ -427,6 +427,11 @@ export function zodToConvex<Z extends zCore.$ZodType>(
   const visited = new WeakSet<zCore.$ZodType>();
 
   function zodToConvexInner(validator: zCore.$ZodType): GenericValidator {
+    // Guard: Ensure validator is an object before using WeakSet
+    if (typeof validator !== "object" || validator === null) {
+      return v.any();
+    }
+    
     // Circular validator definitions are not supported by Convex validators,
     // so we use v.any() when there is a cycle.
     if (visited.has(validator)) {
@@ -517,6 +522,11 @@ export function zodOutputToConvex<Z extends zCore.$ZodType>(
   const visited = new WeakSet<zCore.$ZodType>();
 
   function zodOutputToConvexInner(validator: zCore.$ZodType): GenericValidator {
+    // Guard: Ensure validator is an object before using WeakSet
+    if (typeof validator !== "object" || validator === null) {
+      return v.any();
+    }
+    
     // Circular validator definitions are not supported by Convex validators,
     // so we use v.any() when there is a cycle.
     if (visited.has(validator)) {


packages/convex-helpers/server/zod4.ts
```diff
@@ -427,6 +427,11 @@ export function zodToConvex<Z extends zCore.$ZodType>(
   const visited = new WeakSet<zCore.$ZodType>();
 
   function zodToConvexInner(validator: zCore.$ZodType): GenericValidator {
+    // Guard: Ensure validator is an object before using WeakSet
+    if (typeof validator !== "object" || validator === null) {
+      return v.any();
```
ianmacartney, line 432: why wouldn't this be an exception? Are we hitting this b/c of circular references leading to undefined being passed in? Seems safer to throw an exception here

PR #896, opened 2026-01-10
cut down on per-operation overhead for triggers

ianmacartney:
When doing 200 inserts, it was taking an extra 600+ms even when the trigger was a no-op, due to fetching the document after. Without the redundant fetch, it was ~12ms

----

## Summary by CodeRabbit

* **Bug Fixes**
  * Trigger operations now reliably fetch existing and new records and construct inserted/updated documents so identifiers are preserved and undefined fields are omitted.

* **Documentation**
  * Added changelog entry describing an optimization to trigger behavior (tradeoff: creation-time in insert callbacks may be slightly imprecise).

* **Chores**
  * Bumped package version to 0.1.112-alpha.1

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12006,7 +12006,7 @@
     },
     "packages/convex-helpers/dist": {
       "name": "convex-helpers",
-      "version": "0.1.112-alpha.0",
+      "version": "0.1.112-alpha.1",
       "license": "Apache-2.0",
       "bin": {
         "convex-helpers": "bin.cjs"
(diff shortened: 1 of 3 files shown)


ianmacartney:
Yeah one concern here is whether the _creationTime differs from Date.now() - e.g. I think we increment by nanoseconds in order to keep it sorted in insert order.. not sure how big of a deal that would be for a trigger to get the less precise time

ianmacartney:
@coderabbitai take note: you can't modify / mutate `_creationTime` or `_id` via `db.replace` or `db.patch`

ianmacartney:
Thanks to Jordan's optimization, this should no longer be necessary

PR #905, opened 2026-01-29
Jamwt/periodic query

jamwt:
usePeriodicQuery - polling query interface with jitter.

For use on expensive queries on content-heavy sites that get invalidated frequently and do not need real-time or strong consistency with respect to other page elements.

----

## Summary by CodeRabbit

* **New Features**
  * Added a periodic (polling) data hook for non‑real‑time fetching with configurable interval, jitter, manual refresh, refresh state, last‑updated info, and error handling.
  * Added an example UI component demonstrating the hook and a "Refresh Now" control, visible in the app.

* **Documentation**
  * Added a comprehensive "Periodic Query (Polling)" guide with examples, options, return values, and a TOC entry.

### DIFF
--- a/packages/convex-helpers/react/usePeriodicQuery.ts
+++ b/packages/convex-helpers/react/usePeriodicQuery.ts
@@ -158,8 +158,6 @@ export function usePeriodicQuery<Query extends FunctionReference<"query">>(
     args === "skip" ? "skip" : JSON.stringify(convexToJson(args as Value));
   const queryName = getFunctionName(query);
 
-  // Track whether we're currently fetching to avoid race conditions
-  const isFetchingRef = useRef(false);
   // Track latest args key to avoid stale updates
   const latestArgsKeyRef = useRef(argsKey);
   useEffect(() => {
@@ -174,10 +172,7 @@ export function usePeriodicQuery<Query extends FunctionReference<"query">>(
 
     try {
       const result = await convex.query(query, args);
-      if (
-        isMountedRef.current &&
-        latestArgsKeyRef.current === argsKey
-      ) {
+      if (isMountedRef.current && latestArgsKeyRef.current === argsKey) {
         setState({
           data: result,
           isRefreshing: false,
@@ -186,10 +181,7 @@ export function usePeriodicQuery<Query extends FunctionReference<"query">>(
         });
       }
     } catch (e) {
-      if (
-        isMountedRef.current &&
-        latestArgsKeyRef.current === argsKey
-      ) {
+      if (isMountedRef.current && latestArgsKeyRef.current === argsKey) {
         setState((s) => ({
           ...s,
           isRefreshing: false,


packages/convex-helpers/react/usePeriodicQuery.ts
```diff
@@ -0,0 +1,267 @@
...
+ * @param base - Base interval in milliseconds
+ * @param jitter - Jitter factor (0-1). Result will be in range [base * (1 - jitter), base * (1 + jitter)]
+ * @returns The next interval in milliseconds
+ */
+function getNextInterval(base: number, jitter: number): number {
+  // Math.random() * 2 - 1 gives a value in range [-1, +1]
+  const jitterMultiplier = Math.random() * 2 - 1;
+  const jitterAmount = base * jitter * jitterMultiplier;
+  return base + jitterAmount;
+}
```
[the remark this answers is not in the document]
jamwt: This is okay. It's really the minimum argument, not the minimum sleep time.

packages/convex-helpers/react/usePeriodicQuery.ts
```diff
@@ -0,0 +1,267 @@
...
+        setState((s) => ({
+          ...s,
+          isRefreshing: false,
+          error: e instanceof Error ? e : new Error(String(e)),
+        }));
+      }
+    } finally {
+      isFetchingRef.current = false;
+    }
+  }, [convex, query, argsKey]);
```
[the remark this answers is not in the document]
jamwt: clever you

packages/convex-helpers/package.json
```diff
@@ -43,6 +43,10 @@
       "types": "./react/sessions.d.ts",
       "default": "./react/sessions.js"
     },
+    "./react/usePeriodicQuery": {
+      "types": "./react/usePeriodicQuery.d.ts",
+      "default": "./react/usePeriodicQuery.js"
+    },
```
ianmacartney, line 49: it will be exported from both here, and from "convex-helpers/react" b/c of the export below - we could probably do just this one since it's not generally something we want to push?

packages/convex-helpers/react/usePeriodicQuery.ts
```diff
@@ -0,0 +1,272 @@
...
```
ianmacartney, line 194: The query is a FunctionReference which unfortunately is not stable, but the queryName is. I put this change in my PR #906

packages/convex-helpers/react/usePeriodicQuery.ts
```diff
@@ -0,0 +1,272 @@
...
+/**
+ * Return type for usePeriodicQuery.
+ */
+export type UsePeriodicQueryResult<T> = {
+  /** Query result, or undefined if never successfully loaded */
+  data: T | undefined;
+  /** True during any fetch (including initial load) */
+  isRefreshing: boolean;
+  /** Timestamp of the last successful fetch, or undefined if never loaded */
+  lastUpdated: Date | undefined;
```
ianmacartney, line 67: Generally Convex uses `number` instead of `Date` in APIs, but I don't feel strongly here

PR #907, opened 2026-01-31
fix: prevent 'Cannot use gte on field undefined' error in splitRange

sethconvex:
When using eq() with undefined on a compound index followed by pagination,
the splitRange function would attempt to access indexFields[i] beyond the
array bounds. This occurred because the commonPrefix loop slices indexFields
while processing equality constraints, but makeCompare and the middleRange
code didn't check array bounds before accessing indexFields[i].

This fix adds bounds checking in:
1. makeCompare loop and condition (lines 85, 88)
2. middleRange construction (line 116)

Also adds a regression test for pagination with eq(undefined) on a
compound index to prevent this issue from recurring.

Fixes the error: 'Cannot use gte on field "undefined"'

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed pagination errors that occurred when querying with certain compound index configurations and equality filters.

* **Tests**
  * Added regression test for pagination with compound indexes to prevent future regressions.

### DIFF
--- a/packages/convex-helpers/server/stream.test.ts
+++ b/packages/convex-helpers/server/stream.test.ts
@@ -815,43 +815,43 @@ describe("stream", () => {
   });
 
   test("pagination with eq(undefined) on compound index", async () => {
-      // Regression test for bug where eq(field, undefined) on a compound index
-      // would cause "Cannot use gte on field 'undefined'" error during pagination
-      const testSchema = defineSchema({
-        items: defineTable({
-          softDeletedAt: v.optional(v.number()),
-          updatedAt: v.number(),
-        }).index("by_active_updated", ["softDeletedAt", "updatedAt"]),
-      });
-      const t = convexTest(testSchema, modules);
-      await t.run(async (ctx) => {
-        // Insert some active (non-deleted) items
-        await ctx.db.insert("items", { updatedAt: 1 });
-        await ctx.db.insert("items", { updatedAt: 2 });
-        await ctx.db.insert("items", { updatedAt: 3 });
-        // Insert some soft-deleted items
-        await ctx.db.insert("items", { softDeletedAt: 100, updatedAt: 4 });
-
-        const query = stream(ctx.db, testSchema)
-          .query("items")
-          .withIndex("by_active_updated", (q) => q.eq("softDeletedAt", undefined))
-          .order("desc");
-
-        // This should not throw "Cannot use gte on field 'undefined'"
-        const result = await query.paginate({ numItems: 2, cursor: null });
-        expect(result.page.length).toBe(2);
-        expect(result.page[0]!.updatedAt).toBe(3);
-        expect(result.page[1]!.updatedAt).toBe(2);
-        expect(result.isDone).toBe(false);
-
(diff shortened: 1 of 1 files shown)


sethconvex:
* **#907** <a href="https://app.graphite.com/github/pr/get-convex/convex-helpers/907?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/get-convex/convex-helpers/907?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ianmacartney:
should I review even though it's a draft?

sethconvex:
yes please! I was hoping to get the prettier pushed but haven't had a second to do that yet

packages/convex-helpers/server/stream.ts
```diff
@@ -82,10 +82,10 @@ function splitRange(
   ) => {
     const range = commonPrefix.slice();
     let i = 0;
-    for (; i < key.length - 1; i++) {
+    for (; i < key.length - 1 && i < indexFields.length; i++) {
```
ianmacartney, line 85: When would key be longer than indexFields? That seems like a bug - unless they indexFields don't include _creationTime, _id, which would be important to respect here. If we want to add more safety let's assert that it's the case, vs. not constraining the range as much as expected

packages/convex-helpers/server/stream.test.ts
```diff
@@ -813,4 +813,45 @@ describe("stream", () => {
       ]);
     });
   });
+
+  test("pagination with eq(undefined) on compound index", async () => {
```
ianmacartney, line 817: This test passes without the proposed changes, so I think we need another test that previously failed

packages/convex-helpers/server/stream.ts
```diff
@@ -111,12 +111,15 @@ function splitRange(
     middleRange = makeCompare(startBoundType, startBound);
   } else if (startBound.length === 0) {
     middleRange = makeCompare(endBoundType, endBound);
-  } else {
+  } else if (indexFields.length > 0) {
```
ianmacartney, line 114: Again, not sure how this could happen, so would rather assert - not sure how we would get here

packages/convex-helpers/server/stream.ts
```diff
@@ -111,12 +111,15 @@ function splitRange(
...
     const startValue = startBound[0]!;
     const endValue = endBound[0]!;
     middleRange = commonPrefix.slice();
     middleRange.push([startBoundType, indexFields[0]!, startValue]);
     middleRange.push([endBoundType, indexFields[0]!, endValue]);
+  } else {
```
ianmacartney, line 120: Would love to have some unit tests just on splitRange to validate that it's doing what it's supposed to for the four variants we expect here -

ianmacartney (changes requested):

packages/convex-helpers/server/stream.test.ts
sethconvex: :+1:

PR #910, opened 2026-02-10
Fix zid registry false positives from Zod 4 parent inheritance

mikecann:
## Summary

- Move the `_zidRegistry` lookup in `zodToConvexCommon` into the `$ZodCustom` branch (the type `zid()` produces), rather than checking it at the top of the function before any type-specific `instanceof` checks. This ensures type-specific handlers like `$ZodString`, `$ZodNumber`, etc. always take priority.
- This is a defensive improvement: since `zid()` always produces a `$ZodCustom`, there's no reason to check the zid registry for other types. Scoping the check to `$ZodCustom` makes the code more precise and prevents any future issues if Zod's internal `_zod.parent` chain were to cause registry lookups to match unexpected types.
- Add regression tests for `zid().describe()` and `z.string().describe()` to ensure derived schemas are handled correctly.

## Test plan
- [x] All 150 zodToConvex tests pass
- [x] All zod4-related tests pass
- [x] Full package test suite passes

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive tests for schema-ID handling, covering derived schemas, parent-chain inheritance, primitives, object fields, and edge cases to prevent regressions.

* **Refactor**
  * Adjusted schema conversion logic to limit registry lookups to the custom-schema path, fixing misidentification across derived schemas.

* **Style**
  * Reflowed several type/generic declarations across packages for consistent formatting (no behavioral changes).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -229,7 +229,6 @@
       "integrity": "sha512-e7jT4DxYvIDLk1ZHmU/m/mB19rex9sv0c2ftBtjSBv+kVM/902eh0fINUzD7UwLLNR+jU585GxUJ8/EBfAM5fw==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.27.1",
         "@babel/generator": "^7.28.5",
@@ -817,7 +816,6 @@
         }
       ],
       "license": "MIT",
-      "peer": true,
       "engines": {
         "node": ">=18"
       },
@@ -841,7 +839,6 @@
         }
       ],
       "license": "MIT",
-      "peer": true,
       "engines": {
         "node": ">=18"
       }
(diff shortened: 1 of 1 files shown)


packages/convex-helpers/react/sessions.ts
```diff
@@ -269,8 +269,9 @@ type SessionMutation<Mutation extends FunctionReference<"mutation">> = (
 ) => Promise<FunctionReturnType<Mutation>>;
 
 // Similar to ReactMutation, but with a sessionId parameter.
-interface ReactSessionMutation<Mutation extends FunctionReference<"mutation">>
```
Nicolapps: These look like reformatting changes that are unrelated to the contents of the PR. Could you please revert them?

mikecann: ye the prettier checks are failing without these changes, opus tells me that these formatting issues are on main too

Nicolapps: You’re right, they seem to be caused by the Prettier update in 50ceecf7071d99daf82f0eda5d2de2d074e83ca0. I’m fixing them on `main`

PR #66, opened 2024-02-22
Move Hono into `convex-helpers` npm package

ianmacartney:
Allows using hono without copy-paste. It is now parameterized by `ActionCtx` when declaring it, instead of importing from a generated file.

PR #913, opened 2026-02-11
feat(validators): respect unknownKeys strip mode in validate()

robelest:
## Summary

- `validate()` now respects `unknownKeys: "strip"` on `v.object()` validators.
- `parse()` union handling now follows strict-first semantics for consistency across Convex repos and with Zod-style union ordering expectations.

## What changed

**`packages/convex-helpers/validators.ts`**
- `validate()` object unknown-field checks now skip rejection when `(validator as any).unknownKeys === "strip"`.
- `stripUnknownFields()` union selection now uses two passes:
  - First matching non-strip member with `allowUnknownFields: false`
  - Otherwise first matching strip object member with `allowUnknownFields: true`
- Removed permissive fallback for non-strip members in `parse()` union stripping.

**`packages/convex-helpers/validators.test.ts`**
- Added coverage for strict-vs-strip union precedence.
- Added coverage for strip-member declaration-order behavior.
- Added regression asserting strict-only unions reject extra fields during `parse()`.

**`packages/convex-helpers/server/validators.test.ts`**
- Updated union parsing tests to explicitly use strip-mode members where unknown-field stripping is expected.

## Notes

- The `convex` npm package (`^1.31.0`) does not yet emit `unknownKeys` in serialized validator JSON, so tests use monkey-patch helpers.
- Combinators like `partial()`, `addFieldsToValidator()`, and `vRequired()` currently lose `unknownKeys` when reconstructing via `v.object(fields)`; out of scope here.

## Related

- `get-convex/convex-backend` PR #348
- `get-convex/convex-test` PR #66

---

## Summary by CodeRabbit

* **New Features**
  * Per-validator "strip unknown fields" behavior updated so parsing can remove unexpected fields while strict validation still rejects them; union handling now respects strip vs strict member semantics and member ordering.

* **Tests**
  * Expanded coverage for strip-mode parsing: objects, unions, nested structures, arrays, matching order effects, strict vs strip interactions, and related error cases.

### DIFF
--- a/packages/convex-helpers/server/validators.test.ts
+++ b/packages/convex-helpers/server/validators.test.ts
@@ -343,6 +343,7 @@ test("validators disallow things when they're wrong", async () => {
 
 describe("validate", () => {
   function withStripUnknownKeys(validator: ReturnType<typeof v.object>) {
+    // TODO: Remove once the Convex SDK exposes unknownKeys in validator JSON.
     (validator as any).unknownKeys = "strip";
     return validator;
   }
@@ -609,6 +610,19 @@ describe("validate", () => {
     expect(result).toEqual({ name: "Alice", age: 30 });
   });
 
+  test("parse strips unknown fields from unions", () => {
+    const validator = v.union(
+      v.object({ name: v.string() }),
+      v.object({ age: v.number() }),
+    );
+    const result = parse(validator, {
+      name: "Alice",
+      age: 30,
+      unknown: "field",
+    });
+    expect(result).toEqual({ name: "Alice" });
+  });
+
   test("parse strips unknown fields from strip-mode unions", () => {
     const validator = v.union(
       withStripUnknownKeys(v.object({ name: v.string() })),
@@ -642,6 +656,39 @@ describe("validate", () => {
   });
 
   test("parse strips unknown fields from nested objects", () => {
+    const validator = v.object({
+      name: v.string(),
+      age: v.number(),
+      details: v.object({
+        name: v.string(),
+        age: v.number(),
+      }),
+      union: v.union(
+        v.object({ name: v.string() }),
+        v.object({ age: v.number() }),
+      ),
+      array: v.array(v.object({ name: v.string() })),
+      record: vv.record(v.string(), v.object({ name: v.string() })),
+    });
+    const result = parse(validator, {
+      name: "Alice",
+      age: 30,
+      details: { name: "Alice", age: 30 },
+      union: { name: "Alice", foo: "bar" },
+      array: [{ name: "Alice", foo: "bar" }],
+      record: { a: { name: "Alice", foo: "bar" } },
+    });
+    expect(result).toEqual({
+      name: "Alice",
+      age: 30,
+      details: { name: "Alice", age: 30 },
+      union: { name: "Alice" },
+      array: [{ name: "Alice" }],
+      record: { a: { name: "Alice" } },
+    });
+  });
+
+  test("parse strips unknown fields from nested objects with strip-mode union members", () => {
     const validator = v.object({
       name: v.string(),
       age: v.number(),
@@ -694,6 +741,19 @@ describe("validate", () => {
     expect(result3).toEqual(undefined);
     const result4 = parse(validator, {});
     expect(result4).toEqual({});
+  });
+
+  test("union matches first member with unknown fields", () => {
+    const validator = v.union(
+      v.object({ name: v.string() }),
+      v.object({ name: v.string(), age: v.number() }),
+    );
+    const result = parse(validator, {
+      name: "Alice",
+      age: 30,
+      unknown: "field",
+    });
+    expect(result).toEqual({ name: "Alice" });
   });
 
   test("union with strip members matches first member with unknown fields", () => {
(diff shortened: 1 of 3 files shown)


ianmacartney:
What happens if I pass this object:
```json
{
  "a": 1,
  "b": 2,
  "c": 3,
}
```
to this validator:
```ts
v.union(
  v.object({
    a: v.number(),    
  }, { unknownFields: "strip" }),
  v.object({
    b: v.number(),
    c: v.optional(v.number()),
  }, { unknownFields: "strip" }),
  v.object({
    a: v.union(v.string(), v.number()),
    b: v.optional(v.number()),
    c: v.any(),
  }, { unknownFields: "strip" }),
)
```
? As a user I might expect to get { a: 1, b: 2, c: 3 } since it was possible, even though the first two technically matched.
Unfortunately we might need to rely on some heuristics here:

1. What captures the most fields / strips the fewest number of fields
2. Biasing towards "strict" objects.
3. What is the most specific type

What have you thought about so far @robelest ?

ianmacartney:
FYI you need to put the PR template CLA disclaimer in the PR description to pass that test

robelest:
@ianmacartney what do you think of something like this for the 3 prs

```
given value V and union members [M1, M2, ... Mn]:

best_strict = None        # first strict member that validates
best_strip  = None        # strip member with highest recognized-key count
best_score  = -1

for each member M:
    if M does not validate V: skip

    if M is strict:
        → return immediately (no stripping needed — it's a perfect match)

    if M is strip:
        score = count of V's keys that exist in M's fields
        if score > best_score:
            best_strip = M
            best_score = score

strip using best_strip (or no-op if none matched)
```

ianmacartney:
High level seems reasonable - though not sure how much time /overhead will be spent here. 
Can you check what Zod does in these cases? Might be worth having consistent behavior with ours - especially since zod validators are sometimes used alongside.
The simplest approach would be "the first strict one that matches, otherwise the first stripped one that matches" - and then it's on the developer to order their validators

packages/convex-helpers/validators.ts
```diff
@@ -850,6 +853,32 @@ export function parse<T extends Validator<any, any, any>>(
...
+  );
+}
+
+function firstMatchingUnionMember(
+  members: Validator<any, any, any>[],
+  value: unknown,
+  opts: { allowUnknownFields: boolean; includeStripObjects: boolean },
+): Validator<any, any, any> | undefined {
+  for (const member of members) {
+    if (isStripObjectValidator(member) !== opts.includeStripObjects) {
```
ianmacartney:
```suggestion
    if (isStripObjectValidator(member) &&  !opts.includeStripObjects) {
```

packages/convex-helpers/server/validators.test.ts
```diff
@@ -604,10 +609,10 @@ describe("validate", () => {
     expect(result).toEqual({ name: "Alice", age: 30 });
   });
 
-  test("parse strips unknown fields from unions", () => {
+  test("parse strips unknown fields from strip-mode unions", () => {
     const validator = v.union(
-      v.object({ name: v.string() }),
-      v.object({ age: v.number() }),
```
ianmacartney, line 610: let's duplicate these tests for the strip variants instead of changing the existing ones

packages/convex-helpers/validators.ts
```diff
@@ -883,18 +912,22 @@ function stripUnknownFields<T extends Validator<any, any, any>>(
...
+
+      const stripMember = firstMatchingUnionMember(validator.members, value, {
+        allowUnknownFields: true,
+        includeStripObjects: true,
+      });
+      if (stripMember) {
+        return stripUnknownFields(stripMember, value);
       }
+
       throw new Error("No matching member in union");
```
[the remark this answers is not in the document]
ianmacartney: I think you misunderstand - the first pass should be strict and the second pass permissive - strict meaning it doesn't allow unknown fields. Now that the second pass also checks strict objects, it will allow for strict objects with extra fields to be stripped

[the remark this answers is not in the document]
ianmacartney: Correct. the proposed change is
```
- if (isStripObjectValidator(member) !== opts.includeStripObjects) {
+ if (isStripObjectValidator(member) &&  !opts.includeStripObjects) {
```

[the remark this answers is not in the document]
robelest: Addressed in 041d575. The union parse path now does exactly the strict-then-permissive behavior Ian requested:
- pass 1: strict (no unknown fields), skipping strip objects
- pass 2: permissive (`allowUnknownFields: true`) across both strict and strip object members
This restores permissive stripping for strict-only unions with extra fields while keeping the strict-first preference.

packages/convex-helpers/validators.ts
robelest: Applied exactly in 041d575. `firstMatchingUnionMember` now uses:

```ts
if (isStripObjectValidator(member) && !opts.includeStripObjects) {
  continue;
}
```

So strict pass excludes strip objects, while permissive pass includes both strict and strip members.

packages/convex-helpers/server/validators.test.ts
robelest: Done in 041d575. I restored the original baseline tests and added separate strip-mode variants instead of replacing them:
- `parse strips unknown fields from unions` + `parse strips unknown fields from strip-mode unions`
- `parse strips unknown fields from nested objects` + strip-mode nested-union variant
- `union matches first member with unknown fields` + strip-mode variant

PR #950, opened 2026-04-09
fix(cors): address bugs in allow all origin handling

erquhart:
- append Vary header instead of setting, respect user headers
- no Vary header change when allowed origin is *
- allow all origins + allow credentials is a cors no-op, neither header goes back in that case
- send literal * for allow all origins instead of reflecting request origin

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected CORS header behavior: wildcard origins no longer allow credentials; Access-Control-Allow-Origin, Access-Control-Allow-Credentials and Vary headers are now set conditionally per response to follow CORS rules.

* **Tests**
  * Expanded CORS test coverage with new scenarios and test routes validating origin-specific responses, preflight handling, and appending Origin to existing Vary headers.

* **Documentation**
  * Clarified inline docs around allow-credentials and wildcard origins.

### DIFF
--- a/packages/convex-helpers/server/cors.test.http.ts
+++ b/packages/convex-helpers/server/cors.test.http.ts
@@ -198,4 +198,21 @@ cors.route({
   allowedOrigins: ["http://localhost:3000"],
 });
 
+/**
+ * Test that Vary: Origin is appended to existing Vary headers.
+ */
+cors.route({
+  path: "/existingVaryHeader",
+  method: "GET",
+  handler: httpActionGeneric(async () => {
+    return new Response(JSON.stringify({ message: "has vary" }), {
+      headers: {
+        "Content-Type": "application/json",
+        Vary: "Accept-Encoding",
+      },
+    });
+  }),
+  allowedOrigins: ["http://localhost:3000"],
+});
+
 export default http;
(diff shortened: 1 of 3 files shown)


ianmacartney (commented):
If we're going to possibly break folks, it should throw at import time, not no-op at runtime - that seems like a hard thing to debug. 
Some people might deem it OK to allow requests from any origin, provided that they're not using cookies but instead just explicit headers that the client would need to already have access to. 
I agree we should tighten things up, but ideally it would be:

1. A type error - if you set allowCredentials, then allowedOrigins is required
2. Import time exception - if you set both, it should throw (since it won't work at runtime)
3. Escape-hatch-able

erquhart:
@ianmacartney the break is for folks who use wildcard allow origin while allow credentials is true, which is already a cors error in browsers - this is just matching that behavior. I've added a thrown error at deploy time but can add types in too if preferred. Did you see any other breaks?

Edit: ah we were returning specific origins so this was not a cors error. Taking another look.

erquhart:
Going to remove the wildcard + allow all origins blocking and just post a warning for now, indeed folks might intentionally have it set that way currently.

erquhart:
Closing this, the fact that we reflect the origin instead of returning a literal wildcard makes all of this potentially breaking, can revisit if we want to spend the time in the future.

PR #958, opened 2026-04-17
Fix open-api-spec crash on nullable fields with empty schema

### LINKED ISSUE #957: open-api-spec: SyntaxError on nullable fields with empty schema
## Description

`convex-helpers open-api-spec` fails with a YAML syntax error when the Convex schema includes optional/nullable fields that produce empty JSON Schema objects.

## Error

```
Failed to generate TypeScript API spec:  SyntaxError: All collection items must start at the same column (26728:15)
  26726 |               - updatedAt
  26727 |             properties:
> 26728 |               claimedAt:
        |               ^^^^^^^^^^
> 26729 |                 {}
        | ^^^^^^^^^^^^^^^^^^
> 26730 |                 nullable: true
```

## Cause

In `generateSchemaFromValidator`, the `union` case for nullable types (line ~4509 in `bin.cjs`) concatenates the inner schema with `nullable: true`:

```js
if (nonNullMembers.length === 1 && nullMember !== void 0) {
  return `${generateSchemaFromValidator(nonNullMembers[0])}
nullable: true`;
}
```

When the non-null member is type `"any"`, `generateSchemaFromValidator` returns `"{}"`. The output becomes:

```yaml
claimedAt:
  {}
  nullable: true
```

Prettier's YAML parser (called at line ~4429) rejects this as invalid indentation — `{}` and `nullable: true` appear as sibling collection items at inconsistent columns.

## Fix

When the inner schema is an empty object (`{}`), emit only `nullable: true`:

```diff
      if (nonNullMembers.length === 1 && nullMember !== void 0) {
-        return `${generateSchemaFromValidator(
-          nonNullMembers[0]
-        )}
+        const innerSchema = generateSchemaFromValidator(nonNullMembers[0]);
+        if (innerSchema === "{}") {
+          return "nullable: true";
+        }
+        return `${innerSchema}
 nullable: true`;
      }
```

## Reproduction

Any Convex schema with an optional timestamp or nullable field:

```ts
import { convexTable, timestamp } from "better-convex/orm";

export const example = convexTable("example", {
  claimedAt: timestamp(), // nullable — no .notNull()
});
```

```bash
npx convex-helpers open-api-spec
```

## Environment

- convex-helpers: 0.1.112, 0.1.114 (both affected)
- prettier: 3.8.1
- convex: 1.31.7

ianmacartney:
When a nullable union contained an "any" type, generateSchemaFromValidator
produced "{}\nnullable: true" which is invalid YAML. Since an empty schema
({}) already permits any value including null in OpenAPI 3.0, we now return
"{}" unchanged instead of appending nullable: true.

Fixes #957

### DIFF
--- a/packages/convex-helpers/cli/openApiSpec.test.ts
+++ b/packages/convex-helpers/cli/openApiSpec.test.ts
@@ -29,3 +29,43 @@ test("generateValidSpec", async () => {
 
   expect(JSON.parse(output.toString())["totals"]).toHaveProperty("errors", 0);
 }, 10000);
+
+// Regression test: nullable fields with "any" type (empty schema {}) should produce valid YAML.
+test("nullable any field generates valid spec", async () => {
+  const functionsJson = JSON.stringify({
+    url: "https://test-convex-url.convex.cloud",
+    functions: [
+      {
+        args: {
+          type: "object",
+          value: {
+            claimedAt: {
+              fieldType: {
+                type: "union",
+                value: [{ type: "any" }, { type: "null" }],
+              },
+              optional: false,
+            },
+          },
+        },
+        functionType: "Query",
+        identifier: "example.js:get",
+        returns: { type: "any" },
+        visibility: { kind: "public" },
+      },
+    ],
+  });
+
+  const apiSpec = generateOpenApiSpec(JSON.parse(functionsJson), true);
+
+  const testFileName = "openApiSpec.nullable-any.test.yaml";
+  fs.writeFileSync(testFileName, apiSpec, "utf-8");
+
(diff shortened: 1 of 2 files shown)


ianmacartney:
* **#958** <a href="https://app.graphite.com/github/pr/get-convex/convex-helpers/958?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/get-convex/convex-helpers/958?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

jordanhunt22 (approved):
awesome

PR #962, opened 2026-04-21
Streams: make page split recommendations less aggressive

dowski:
Previously a split was recommended when the number of scanned documents was 1 greater than the number of requested documents. For queries where many documents had to be scanned to satisfy the requested number of documents, this would lead to cascading page splits.

Now split recommendations based on number of documents scanned use a threshold based on the published limits. A split recommendation will also happen if the returned page size is somewhat more than the requested page size, which can only happen with an end cursor set.

If there is no end cursor, the page results are naturally truncated by the requested number of items (that's existing behavior).

This addresses https://github.com/get-convex/convex-backend/issues/443.

### DIFF
--- a/packages/convex-helpers/server/stream.ts
+++ b/packages/convex-helpers/server/stream.ts
@@ -428,8 +428,15 @@ export abstract class QueryStream<
       pageStatus = "SplitRequired";
       splitCursor = indexKeys[Math.floor((indexKeys.length - 1) / 2)];
     } else if (indexKeys.length >= softMaxRowsToRead) {
-      pageStatus = "SplitRecommended";
-      splitCursor = indexKeys[Math.floor((indexKeys.length - 1) / 2)];
+      // When the scan-to-match ratio is high (sparse filter), splitting the
+      // range won't improve density — it just doubles subscriptions and can
+      // cause exponential split cascades on the client.
+      const scanRatio =
+        page.length > 0 ? indexKeys.length / page.length : Infinity;
+      if (scanRatio <= 2) {
+        pageStatus = "SplitRecommended";
+        splitCursor = indexKeys[Math.floor((indexKeys.length - 1) / 2)];
+      }
     }
     return {
       page,
--- a/packages/convex-helpers/server/stream.ts
+++ b/packages/convex-helpers/server/stream.ts
@@ -30,9 +30,13 @@ import type {
 export type IndexKey = (Value | undefined)[];
 
 // From https://docs.convex.dev/production/state/limits#transactions
-const MAX_DOCUMENTS_TO_SCAN = 32000;
-// Value used to trigger page split suggestions (nearing the max).
-const SOFT_MAX_SCAN_LEN = (MAX_DOCUMENTS_TO_SCAN * 3) / 4;
+const MAX_DOCUMENT_SCAN_LEN = 32000;
+// Value used to trigger page split suggestions. This is kind of
+// a backstop value to mitigate additional scans over the full
+// original range if documents change later. Making it smaller
+// would result in more splits (equal to the divisor) as the
+// number of documents scanned approaches the max.
+const SOFT_MAX_SCAN_LEN = MAX_DOCUMENT_SCAN_LEN / 2;
 
 //
 // Helper functions
--- a/packages/convex-helpers/server/stream.ts
+++ b/packages/convex-helpers/server/stream.ts
@@ -29,6 +29,11 @@ import type {
 
 export type IndexKey = (Value | undefined)[];
 
+// From https://docs.convex.dev/production/state/limits#transactions
+const MAX_DOCUMENTS_TO_SCAN = 32000;
+// Value used to trigger page split suggestions (nearing the max).
+const SOFT_MAX_SCAN_LEN = (MAX_DOCUMENTS_TO_SCAN * 3) / 4;
+
 //
 // Helper functions
 //
@@ -426,7 +431,10 @@ export abstract class QueryStream<
     if (hitLimit) {
       pageStatus = "SplitRequired";
       splitCursor = indexKeys[Math.floor((indexKeys.length - 1) / 2)];
-    } else if (page.length > opts.numItems + 1) {
+    } else if (
+      indexKeys.length >= SOFT_MAX_SCAN_LEN ||
+      page.length > opts.numItems + 1
+    ) {
       pageStatus = "SplitRecommended";
       splitCursor = indexKeys[Math.floor((indexKeys.length - 1) / 2)];
     }
--- a/packages/convex-helpers/server/stream.ts
+++ b/packages/convex-helpers/server/stream.ts
@@ -29,6 +29,11 @@ import type {
 
 export type IndexKey = (Value | undefined)[];
 
+// From https://docs.convex.dev/production/state/limits#documents
+const MAX_ARRAY_LEN = 8192;
+// Value used to trigger page split suggestions (nearing the max).
+const SOFT_MAX_PAGE_LEN = (MAX_ARRAY_LEN * 3) / 4;
+
 //
 // Helper functions
 //
@@ -374,7 +379,6 @@ export abstract class QueryStream<
       inclusive: true,
     };
     const maxRowsToRead = opts.maximumRowsRead;
-    const softMaxRowsToRead = opts.numItems + 1;
     let maxRows: number | undefined = opts.numItems;
     if (opts.endCursor) {
       newEndKey = {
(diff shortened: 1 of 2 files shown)


ianmacartney:
Splitting still makes sense to me for sparse pages - if I'm scanning 1k records and only getting 10 results, I don't want to scan 1.1k records to get the 11th result, I'd rather only scan 5k.
I see splitting as reducing the read bandwidth per update, vs. the results size. 
But I guess over-fetching on one request then re-fetching that same page recursively many times doesn't help.

It makes me think that often the split cursor you want isn't always at the middle, and that splitting maybe should only happen when a page changes size- if you're scanning a bunch of static records, it doesn't help. It helps when something is growing

dowski:
That's fair. Another approach we could take is limiting the splitting depth on the client.

Imagine you request page A and get a split suggestion for B and C. Then you fetch those and get split suggestions for D, E, F and G. We could track that the lineage of the split was from the initial request A and stop splitting at a certain depth.

Maybe that's something we expose to callers since its effectiveness will depend on the underlying dataset and the type of filtering that an app does.

ianmacartney:
Or if it could provide all of the split cursors at once, then we only have
one refresh - where 1 query turns into N sub-queries, reading the data
twice but not D times (D = depth of splits)

ianmacartney:
I think regardless the big win here would be detecting change / refreshes -
avoiding multiplying bandwidth when pages are static. And otherwise if
we're going to double, let's get to a page size that won't be super
expensive when things change. Avoiding splitting only helps if the page
isn't changing - if something is fetching hundreds of items and doesn't
split, then each change will re-fetch hundreds of items.

ianmacartney:
@dowski have you given this more thought? I think the "only split if it changes" heuristic feels right to me:  
1\. If the data isn't changing, they paid the big scan cost only once  
2\. If the data on one end of the range is changing, they re-fetch that half, and on each change they'll be fetching half as many duplicates - for a page that was originally scanning 800 items that updates on the end they might eventually have `1-400 | 400-600 | 600-700 | 700-750 | <smaller page more frequently updating>` - so log(N) refetches.

dowski:
That seems like a cool idea. I wonder how we'd implement it though. It seems like we'd need some indication in the query function that it is executing based on a fresh subscription from a client or a re-execution due to data change.

Maybe that gets exposed in `ctx.meta.getFunctionMetadata()`? Then it could be passed into helper code like `stream.paginate` which would inspect it and condition split behavior like you described.

```typescript

type PaginationOptsWithQuerySource = PaginationOpts & {querySource: "client" | "system"};

export const listForAuthors = query({
  args: {
    authors: v.array(v.id("users")),
    paginationOpts: paginationOptsValidator,
  },
  handler: async (ctx, { authors, paginationOpts }) => {
    const authorStreams = authors.map((author) =>
      stream(ctx.db, schema)
        .query("messages")
        .withIndex("by_author", (q) => q.eq("author", author)),
    );
    const allAuthorsStream = mergedStream(authorStreams, [
      "author",
      "_creationTime",
    ]);
    // Paginate the result differently depending on whether the source of the query execution is a client
    // subscription request or a system initiated re-execution.
    return await allAuthorsStream.paginate({
        ...paginationOpts,
        querySource: ctx.meta.getFunctionMetadata().querySource,
    });
  },
});
```

dowski:
> Maybe that gets exposed in `ctx.meta.getFunctionMetadata()`?

`ctx.meta.getFunctionMetadata()` is probably not be the exact right spot, but I think the general idea is the same.

ianmacartney (changes requested):
going to put back in your queue for now

packages/convex-helpers/server/stream.ts
```diff
@@ -428,8 +428,15 @@ export abstract class QueryStream<
       pageStatus = "SplitRequired";
       splitCursor = indexKeys[Math.floor((indexKeys.length - 1) / 2)];
     } else if (indexKeys.length >= softMaxRowsToRead) {
```
ianmacartney: I think we want to compare how many of these rows got "used" - not sure if we have that, or if the filterWith layering is elsewhere. also only if endCursor is set, per our chat the other day

dowski: I think this is addressed now.

packages/convex-helpers/server/stream.ts
```diff
@@ -427,7 +431,14 @@ export abstract class QueryStream<
     if (hitLimit) {
       pageStatus = "SplitRequired";
       splitCursor = indexKeys[Math.floor((indexKeys.length - 1) / 2)];
-    } else if (indexKeys.length >= softMaxRowsToRead) {
+    } else if (
+      (indexKeys.length >= SOFT_MAX_PAGE_LEN ||
+        page.length >= opts.numItems * 2) &&
```
ianmacartney:
```suggestion
        page.length > opts.numItems) &&
        page.length > 4 &&
```
Call me crazy, but I still believe in eagerly splitting - this should only happen on changes/inserts, vs. automatically recursively.
The client tries to eagerly split, but won't be able to if there isn't a splitCursor returned from the server.

If you want to keep it as a multiple, how about 1.5x? but I think part of what makes the helpers pagination preferable for bandwidth (users have commented on this), is the shifting of average bandwidth from `(n + m/2) * m` to `n + (n/2 + m/2) * m` for `m` updates to a page size of `n` (m < n).

packages/convex-helpers/server/stream.ts
```diff
@@ -29,6 +29,11 @@ import type {
 
 export type IndexKey = (Value | undefined)[];
 
+// From https://docs.convex.dev/production/state/limits#documents
+const MAX_ARRAY_LEN = 8192;
+// Value used to trigger page split suggestions (nearing the max).
+const SOFT_MAX_PAGE_LEN = (MAX_ARRAY_LEN * 3) / 4;
```
ianmacartney: worth noting that they could technically scan up to 32k documents - we don't want the results to be > 8192 since we can't to return that from the query, but scanning more I think wouldn't technically break anything.

I'm ok with this limit - if they're scanning that many, they are already unable to do a join on all those entries (only 4096 db queries allowed in a tx). 

Worst case they have a query scanning 32k documents and it's auto-split into 16k/16k, then 8k/8k,8k/8k, then 4k/4k,4k/4k,4k/4k,4k/4k - tripling the bandwidth...

Another option here would be to set the soft limit to `3/4*maxRowsToRead` and set the default for `maxRowsToRead` to 8192 items, so by default users don't ever have their pages auto-splitting recursively (would only split once), and users who want to scan more (up to 32k) could set `maxRowsToRead` to the higher number and it would still only split once.

packages/convex-helpers/server/stream.ts
```diff
@@ -427,7 +431,14 @@ export abstract class QueryStream<
     if (hitLimit) {
       pageStatus = "SplitRequired";
       splitCursor = indexKeys[Math.floor((indexKeys.length - 1) / 2)];
-    } else if (indexKeys.length >= softMaxRowsToRead) {
+    } else if (
+      (indexKeys.length >= SOFT_MAX_PAGE_LEN ||
+        page.length >= opts.numItems * 2) &&
+      opts.endCursor
```
ianmacartney: if endCursor isn't set, I don't think we could end up with `page.length > opts.numItems`, so I think this is fine, but also always satisfied?

packages/convex-helpers/server/stream.ts
```diff
@@ -427,7 +431,14 @@ export abstract class QueryStream<
...
+    ) {
+      // Recommend a split when an end cursor is set and either of the following is true of the page:
+      // 1. It is approaching transaction limits (the "soft max" check).
+      // 2. It contains at least twice the number of documents that the client initially requested.
       pageStatus = "SplitRecommended";
       splitCursor = indexKeys[Math.floor((indexKeys.length - 1) / 2)];
```
ianmacartney, line 443: if we go with 1.5x, one heuristic here would be to have the split cursor be at 1/3 of the way in when cursor is null - assume that the growth is at the start of the range. 
While it might seem like a small change - it could mean the growth looks like 
(initial page of 4, split at 1.5x at 1/3)
```
+                         total
  [1, 2, 3, 4]              4
1 [1, 2, 3, 4, 5]           9
2 [1, 2, 3, 4, 5, 6]        15
* [1, 2, 3, 4],[5, 6]       21
3 [5, 6, 7]                 24
4 [5, 6, 7, 8]              28
5 [5, 6, 7, 8, 9]           32
```
instead of splitting at 2x at 1/2
```
+                         total
  [1, 2, 3, 4]              4
1 [1, 2, 3, 4, 5]           9
2 [1, 2, 3, 4, 5, 6]        15
3 [1, 2, 3, 4, 5, 6, 7]     22
4 [1, 2, 3, 4, 5, 6, 7, 8]  30
* [1, 2, 3, 4],[5, 6, 7, 8] 38
5 [5, 6, 7, 8, 9]           43
```
(34% more!)
The effect is more pronounced the more items are added

ianmacartney (approved):
If you're ok with the suggested change, this LGTM.

I'd also advocate for the soft limit to be `3/4*maxRowsToRead` and setting the default for `maxRowsToRead` to `max(8192, numItems)`, but that might be more controversial, maybe some users are scanning 16k+ records reliably (and never hitting 32k scans and crashing)

packages/convex-helpers/server/stream.ts
dowski: Yeah, I see what you mean and this makes sense.

I'm also changing the scan guard to be more accurate. I think it's important that it remain in place because we were implicitly guarding against large scans in the previous implementation. If drop to just splitting on page length growth then we might regress users who were counting on the scan length prevention without knowing it.

packages/convex-helpers/server/stream.ts
dowski: Good point, removed.

packages/convex-helpers/server/stream.ts
dowski: Nice analysis! I went with your more eager splitting suggestion though.

packages/convex-helpers/server/stream.ts
dowski: Thanks, adjusted the logic based on this.

packages/convex-helpers/server/stream.ts
```diff
@@ -29,6 +29,11 @@ import type {
 
 export type IndexKey = (Value | undefined)[];
 
+// From https://docs.convex.dev/production/state/limits#transactions
+const MAX_DOCUMENTS_TO_SCAN = 32000;
+// Value used to trigger page split suggestions (nearing the max).
+const SOFT_MAX_SCAN_LEN = (MAX_DOCUMENTS_TO_SCAN * 3) / 4;
```
ianmacartney:
```suggestion
const SOFT_MAX_SCAN_LEN = MAX_DOCUMENTS_TO_SCAN / 2;
```
we can keep the normal behavior, but IMO we could be setting a soft limit much lower but /2 still has the property of splitting at most once for a given initial query range (won't go recursive)

ianmacartney (approved):
nice!
This is such a big fix / was so glaring, I hope users start picking this up.

ianmacartney:
convex-helpers@0.1.118

PR #967, opened 2026-05-05
Add lazy split strategy for usePaginatedQuery

dowski:
When paginated queries produce sparse results, the server repeatedly recommends page splits that can be excessive for inactive data ranges. These split recommendations can also recurse, leading to exponential queries. The new `splitStrategy: "lazy"` option defers acting on SplitRecommended until the backend sends updated data for the page, so splits only happen where the data is actively changing.

Default remains "eager" for backwards compatibility. SplitRequired and page-size-based splits always execute immediately regardless of strategy.

This is meant to address the issue raised in https://github.com/get-convex/convex-backend/issues/443.

----

### DIFF
--- a/packages/convex-helpers/react/cache/hooks.ts
+++ b/packages/convex-helpers/react/cache/hooks.ts
@@ -420,6 +420,11 @@ export function usePaginatedQuery<Query extends PaginatedQueryReference>(
   // For lazy split strategy: track result references when SplitRecommended is
   // first seen. A new reference from the backend (even with identical content)
   // means the underlying query was re-evaluated, signaling active data.
+  // TODO: On warm-cache remounts, the client re-delivers cached data as a
+  // new reference, which we can't distinguish from a genuine backend update.
+  // This triggers a previously deferred split despite no update from the
+  // backend. Needs a client-side signal (e.g. a delivery version) to
+  // differentiate cache re-delivery from fresh data.
   const deferredSplitRefs = useRef<
     Record<QueryPageKey, PaginationResult<Value>>
   >({});
--- a/packages/convex-helpers/react/cache/hooks.ts
+++ b/packages/convex-helpers/react/cache/hooks.ts
@@ -19,7 +19,7 @@ import type {
   PaginationResult,
 } from "convex/server";
 import { getFunctionName } from "convex/server";
-import { useContext, useEffect, useMemo, useState } from "react";
+import { useContext, useEffect, useMemo, useRef, useState } from "react";
 import { ConvexQueryCacheContext } from "./provider.js";
 import {
   ConvexError,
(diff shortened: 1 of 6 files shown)


packages/convex-helpers/react/cache/hooks.ts
```diff
@@ -398,6 +415,19 @@ export function usePaginatedQuery<Query extends PaginatedQueryReference>(
   }
   const convexClient = useConvex();
   const logger = convexClient.logger;
+  const splitStrategy = options.splitStrategy ?? "eager";
+
+  // For lazy split strategy: track result references when SplitRecommended is
+  // first seen. A new reference from the backend (even with identical content)
+  // means the underlying query was re-evaluated, signaling active data.
+  // TODO: On warm-cache remounts, the client re-delivers cached data as a
+  // new reference, which we can't distinguish from a genuine backend update.
```
ianmacartney, line 424: can you say more about warm-cache remounts / what's happening here? re-mounting should end up as an initial state, and afaik all the references should be the same when re-mounted if it's still cached

ianmacartney (commented):
Cool! Digging in now. A couple thoughts off the top of my head:

1. It seems like returning SplitRecommended on the initial load is a bug - that we should be comparing how many documents we return, not the number we scan - I'm not sure why the logic diverges from the normal pagination in this way - if anything the intent was to reduce bandwidth, which its obviously failing at currently. I have an idea below*
2. Who benefits from the "eager" default?  Is the only difference for people who are making queries that return SplitRecommended immediately? Could we change the default immediately without negative impact?
3. After we change the split strategy in helpers, we also would need to change the strategy in the normal convex react client hooks too, right? I suspect a lot of users use the normal pagination hooks with helpers - even though it's not recommended, it normally is fine (if the data isn't changing)

*What if we return SplitRecommended/Required only if endCursor is set and either the request page is larger than requested or the max rows / max size limits were hit or the transaction size is getting too big (using more than half of the available transaction limits).

packages/convex-helpers/react/cache/hooks.ts
```diff
@@ -445,27 +476,37 @@ export function usePaginatedQuery<Query extends PaginatedQueryReference>(
...
+            break;
+          case "defer":
+            deferredSplitRefs.current[pageKey] = currResult;
+            break;
+          case "none":
+            // The page no longer has a splitCursor (e.g. the server stopped
+            // recommending a split after data changed), so clean up the ref.
+            if (prevResult !== undefined && !currResult.splitCursor) {
+              delete deferredSplitRefs.current[pageKey];
+            }
```
ianmacartney, line 507: one thing that I think isn't captured here, that will be valuable once we fix the "bug" with SplitRecommended: when the page is going from "no split recommended" to "the page changed and now it's recommended" this logic won't currently do a split, but I think it should. 
The current logic will "defer" since there wasn't a splitcursor present before, so deferredSplitRefs doesn't have an entry. Instead I think whenever the data changes and it has SplitRecommended (whether it was recommending a split before or not) it should split - so we're only really excluding the case where SplitRecommended comes on the first page, which should be in exceptional circumstances once we account for it in the server

ianmacartney (commented):
I'm going to wait for the "lazy-by-default" version to re-review, but it also sounds like we can fix this mostly on the server

PR #978, opened 2026-05-21
Zod 4 support: fix VObject Type slot under exactOptionalPropertyTypes

AtAFork:
For `z.object({ k: z.X().optional() })` the bridge wrote `{ k?: X | undefined }` into VObject's Type slot via `zCore.infer<Z>`, which isn't assignable to Convex's `Value` when downstream code is compiled with `exactOptionalPropertyTypes: true`. Anything expecting `Record<string, Value>` (Doc<>, patch args, helper-typed function args) wouldn't compile for those consumers.

Switch to `ObjectType<F>` at the `$ZodObject` branch and `ConvexObjectValidatorFromRecord`. That's the helper `v.object(...)` itself uses, so bridge output is now structurally identical to a literal v.object call. Brand attachment for `$ZodBranded<$ZodObject>` is preserved via a nested infer; existing brand tests pass unchanged.

Also removes the `z.brand()` outer branch as dead code: under Zod 4's structural model `$ZodBranded<T, B>` is `T & {marker}`, so branded primitives match `$ZodString`/`$ZodNumber`/etc first and branded objects match `$ZodObject` first.

Type-level only, no runtime change, no JS emit diff. Consumers without `exactOptionalPropertyTypes` see structurally equivalent types; consumers with it can now use the bridge. CHANGELOG flags the `.type` slot as strictly tighter.

New regression block covers one/two/four-level nesting, branded object with optional field, literal-keyed record (the actual `ConvexObjectValidatorFromRecord` path), and `.partial()`. Recursive linked-list tests cast through `$ZodType` to dodge eager type expansion on self-referential schemas, same workaround pattern the pre-fix code used via `@ts-expect-error`.

----

### DIFF
--- a/packages/convex-helpers/server/zod4.zodtoconvex.mini.test.ts
+++ b/packages/convex-helpers/server/zod4.zodtoconvex.mini.test.ts
@@ -726,10 +726,11 @@ describe("zodToConvex + zodOutputToConvex", () => {
   });
 
   test("recursive type", () => {
-    // Bypass the `Equals<>`-constrained helper: the self-referential schema
-    // forces the bridge type to expand forever through `ObjectType<F>`.
-    // Compile-time shape verification is covered by the non-recursive tests;
-    // this only asserts the runtime conversion produces the right validator.
+    // Cast through `as zCore.$ZodType` to keep the Type slot opaque: a
+    // self-referential schema makes `ConvexValidatorFromZod` circularly
+    // reference itself, a recoverable TS2615. Compile-time shape verification
+    // is covered by the non-recursive tests; this only asserts the runtime
+    // conversion produces the right validator.
     const category = z.object({
       name: z.string(),
       get subcategories() {
--- a/packages/convex-helpers/server/zod4.zodtoconvex.mini.test.ts
+++ b/packages/convex-helpers/server/zod4.zodtoconvex.mini.test.ts
@@ -738,7 +738,10 @@ describe("zodToConvex + zodOutputToConvex", () => {
       },
     });
 
-    expect(zodToConvex(category as zCore.$ZodType)).to.deep.equal(
+    testZodToConvexInputAndOutput(
+      category,
+      // @ts-expect-error -- the TypeScript type is recursive, while the
+      // runtime time uses `v.any()` since Convex validators can’t be recursive
       v.object({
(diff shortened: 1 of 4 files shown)


packages/convex-helpers/server/zod4.zodtoconvex.mini.test.ts
```diff
@@ -726,16 +726,18 @@ describe("zodToConvex + zodOutputToConvex", () => {
...
       name: z.string(),
       get subcategories() {
         return z.array(category);
       },
     });
 
-    testZodToConvexInputAndOutput(
-      category,
-      // @ts-expect-error -- TypeScript can’t compute the full type and uses `unknown`
+    expect(zodToConvex(category as zCore.$ZodType)).to.deep.equal(
```
Nicolapps: There seems to be a regression here: if the user has a recursive type in their code, without the `as zCore.$ZodType` `zodToConvex(category)` causes a type error:

```
Type of property 'subcategories' circularly references itself in mapped type '{ name: never; subcategories: any; }'.
```

I wonder if this could break existing code that’s using the helpers. Do you know if there is a way to avoid it?

Nicolapps (commented):
Thank you for your PR! Having support for `exactOptionalPropertyTypes` would be great; feel free to modify the default settings used by this repository to enable `exactOptionalPropertyTypes` once it works.

packages/convex-helpers/server/zod4.zodtoconvex.mini.test.ts
AtAFork: Ah yea sorry, should be fixed now. The Type slot now comes from Zod's `infer` instead of `ObjectType<F>` over the recursive record, so the self-reference is still lazy.

Nicolapps (approved):
Thank you for the update!

I adjusted comments in 95f9e00 and made a fix to the `EOPTInfer` type in 54243cd that was necessary for the “two-level: parent of optional children that themselves carry optional grandchildren” test to pass when `exactOptionalPropertyTypes` is enabled.

Nicolapps:
I just published convex-helpers@0.1.119 with the fix!

AtAFork:
really appreciate this @Nicolapps!

PR #991, opened 2026-07-21
cli: add timeout and clear error when the deployment is unreachable (open-api-spec / ts-api-spec)

### LINKED ISSUE #877: `npx convex-helpers open-api-spec --include-internal --output-file convex-spec.yaml` appears to hang forever if it cant contact deployment
Silently hangs forever (or at least many minutes). Found this running convex locally, while having the backend not running.

Jean-EmmanuelP:
Fixes #877

## Problem

`npx convex-helpers open-api-spec` (and `ts-api-spec`) hangs forever with no output when the configured deployment can't be reached — e.g. running against a local deployment while `npx convex dev` isn't running.

## Cause

`getFunctionSpec` in `cli/utils.ts` shells out to `npx convex function-spec` via `spawnSync` with no `timeout`, so if the underlying CLI blocks on the network, the command blocks indefinitely.

## Change

- Add a `timeout` to the `spawnSync` call (default **120s**).
- Expose it as `--timeout <seconds>` on both `open-api-spec` and `ts-api-spec`.
- On timeout, fail with an actionable message (check that `npx convex dev` is running / use `--prod` / raise `--timeout`) instead of hanging silently.
- Also surface other `spawnSync` errors (`result.error` was previously ignored — only `status` was checked).

## Testing

- `npm run typecheck` passes (package).
- `npx vitest run cli`: 8/8 tests pass.
- Repro: with an unreachable deployment, the command now exits after the timeout with the error message instead of hanging.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2205,9 +2205,6 @@
         "arm64"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -2225,9 +2222,6 @@
         "arm64"
       ],
       "dev": true,
-      "libc": [
-        "musl"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -2245,9 +2239,6 @@
         "ppc64"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -2265,9 +2256,6 @@
         "s390x"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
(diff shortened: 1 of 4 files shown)


Jean-EmmanuelP:
Good catch on the `--timeout` parsing — added validation: non-numeric or non-positive values now exit with a clear error instead of producing `NaN`.

ianmacartney:
Sorry for the delay here - we aren't working on this currently but there's a test package linked here you can use for now

package-lock.json
```diff
@@ -2205,9 +2205,6 @@
         "arm64"
       ],
       "dev": true,
-      "libc": [
```
jordanhunt22, line 2208: i'm not sure these `package-lock.json` changes are needed. otherwise, this looks good.

jordanhunt22 (approved):
lgtm

PR #993, opened 2026-07-28
Re-export compareValues from convex/values

reeceyang:
Deprecate and replace compareValues with the compareValues in `convex/values`. We can drop it in a future major release.

---


Nicolapps (approved):
Sounds good! We should probably document the subtle breaking change in the changelog

reeceyang:
Actually, maybe we should just remove it and release 0.2.0 - will wait for Ian to weigh in.

ianmacartney:
Re-exporting works for me, and we can drop in the future if we want. I'd like updating packages to resolve the issue without them manually bumping to 0.2. Agreed we should drop a note in the changelog

PR #1004, opened 2026-09-05
fix validation of system table ids (#1003)

### LINKED ISSUE #1003: when given a v.id('_storage') validator and a real db validator() from convex-helpers/validators will throw an error
When you use a `v.id('_storage')` validator a built-in Convex query/mutation/action will accept this fine but anything which calls the `validator` from the convex-helpers package with a real db object, like [@convex-dev/workflow](https://www.convex.dev/components/workflow) does, it will instead error out with a "**Error: System tables can only be accessed from db.system.normalizeId().**" error.

This is due to the way that validate handles ID
https://github.com/get-convex/convex-helpers/blob/aa7fd7a0111d4066fc564acac6cabe63e57fd2d0/packages/convex-helpers/validators.ts#L678-L689

## Reproduction
I've made a full minimum reproduction that shows that `v.id('_storage')` is fine as an arg validator but throws when you try to call `validate` with a real db or use it in a workflow.
https://github.com/dantman/convex-helpers-storageid-validation-bug-reproduction

ianmacartney:
Fixes #1003

### DIFF
--- a/convex/example.test.ts
+++ b/convex/example.test.ts
@@ -13,6 +13,18 @@ describe("testingExample", () => {
     await t.close();
   });
 
+  test("validates system table ids with a real db", async () => {
+    const storageId = await t.action(api.testingFunctions.storeFile, {});
+    expect(
+      await t.query(api.testingFunctions.validateStorageId, { storageId }),
+    ).toBe(true);
+    expect(
+      await t.query(api.testingFunctions.validateStorageId, {
+        storageId: "not an id",
+      }),
+    ).toBe(false);
+  });
+
   test("counter", async () => {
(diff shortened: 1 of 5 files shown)


ianmacartney:
* **#1004** <a href="https://app.graphite.com/github/pr/get-convex/convex-helpers/1004?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/get-convex/convex-helpers/1004?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

